Issue #18, opened 2017-05-30
Local build with "mvn clean verify" fails


vogella:
If I clone the repo and use "mvn clean verify" I get an error:

[INFO] Unpacking org.objectweb.asm_5.2.0.v20170126-0011...
[INFO] Resolving class path of MavenProject: org.eclipse.bluesky:org.eclipse.bluesky:0.1.0-SNAPSHOT @ /home/vogella/git/eclipse-bluesky/org.eclipse.bluesky/pom.xml
[INFO] Computing target platform for MavenProject: org.eclipse.bluesky:org.eclipse.bluesky.feature:0.1.0-SNAPSHOT @ /home/vogella/git/eclipse-bluesky/org.eclipse.bluesky.feature/pom.xml
[INFO] Resolving dependencies of MavenProject: org.eclipse.bluesky:org.eclipse.bluesky.feature:0.1.0-SNAPSHOT @ /home/vogella/git/eclipse-bluesky/org.eclipse.bluesky.feature/pom.xml
[INFO] Resolving class path of MavenProject: org.eclipse.bluesky:org.eclipse.bluesky.feature:0.1.0-SNAPSHOT @ /home/vogella/git/eclipse-bluesky/org.eclipse.bluesky.feature/pom.xml
[INFO] Computing target platform for MavenProject: org.eclipse.bluesky:repository:0.1.0-SNAPSHOT @ /home/vogella/git/eclipse-bluesky/repository/pom.xml
[INFO] Resolving dependencies of MavenProject: org.eclipse.bluesky:repository:0.1.0-SNAPSHOT @ /home/vogella/git/eclipse-bluesky/repository/pom.xml
[INFO] {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, osgi.arch=x86_64}
[ERROR] Cannot resolve project dependencies:
[ERROR]   Software being installed: repository raw:0.1.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):0.1.0-SNAPSHOT
[ERROR]   Missing requirement: repository raw:0.1.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):0.1.0-SNAPSHOT requires 'com.google.guava [18.0.0.v20161115-1643]' but it could not be found
[ERROR] 
[ERROR] See http://wiki.eclipse.org/Tycho/Dependency_Resolution_Troubleshooting for help.
[...]

PR #19, opened 2017-05-31
Fix build.

akurtakov:
Don't hardcode guava version and move to latest Orbit and platform.


akurtakov:
Should fix #18.

mickaelistria:
This fixes #18

mickaelistria:
Merged, thanks.

PR #27, opened 2017-08-19
JSON Schema support.

angelozerr:
This PR add JSON support. It could be improved by providing an jsonSchemaContribution extension point to avoid hard coding the associations between filrPattern and schema to use.

Schema can be a file too. I think it should be good BlueSky hosts some JSON Schema when it is not possible to download the schema.

The JSON Schema contribution is another sample that bluesky plugin should be splitted in several plugins. Each plugin contribute to associations (javascript, typescript, etc).

### DIFF
--- a/org.eclipse.bluesky/src/org/eclipse/bluesky/JSonLanguageServerInterface.java
+++ b/org.eclipse.bluesky/src/org/eclipse/bluesky/JSonLanguageServerInterface.java
@@ -1,0 +1,39 @@
+/**
+ *  Copyright (c) 2017 Angelo ZERR.
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v1.0
+ *  which accompanies this distribution, and is available at
+ *  http://www.eclipse.org/legal/epl-v10.html
+ *
+ *  Contributors:
+ *  Angelo Zerr <angelo.zerr@gmail.com> - initial API and implementation
+ */
+package org.eclipse.bluesky;
+
+import java.util.List;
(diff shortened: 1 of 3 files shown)


org.eclipse.bluesky/src/org/eclipse/bluesky/JSonLanguageServerInterface.java
```diff
@@ -0,0 +1,33 @@
...
+ * 
+ * <ul>
+ * <li>"json/schemaAssociations" to send a mapping between fileMatch (ex:
+ * package.json) and uri (http://json.schemastore.org/package)</li>
+ * </ul>
+ *
+ */
+public interface JSonLanguageServerInterface extends LanguageServer {
+
+	@JsonNotification("json/schemaAssociations")
```
mickaelistria: So the `json/schemaAssociation` operation is a JSON-RPC operation included in VSCode JSon Language Server, right? If so, can you please add a comment to explain that on this method, ideally with a link to some documentation or code in VSCode.

angelozerr: Yes it that. I have added a comment.

mickaelistria:
I updated to TP, hoping it would solve the build issue. Can you please rebase your commit and push it again?

mickaelistria:
Thank you for this good patch! Rebased and merged as c441d9bb8e223eb19759f6c05a3d5de7e6944c60

angelozerr:
Thanks a lot @mickaelistria !

PR #33, opened 2017-08-25
Configure language configuration for css, html, json, javascript, typescript, yaml.

angelozerr:
This PR configure language configuration for each bluesky languages.  language configuration is managed by tm4e and it supports for the moment `auto close` & `on enter` (ex : generate a new indentation after a bracket but not after some other pattern when enter is done).

 * here a little demo with JSON which doesn't use language configuration

jsonwithoutlanguageconfiguration

 * here a little demo with JSON with  language configuration:

jsonwithlanguageconfiguration

Signed-off-by: angelozerr <angelo.zerr@gmail.com>


angelozerr:
I will recreate a new PR.

PR #40, opened 2017-09-04
Enable script and style validation for HTML server

mniewrzal:


mickaelistria:
Thanks for this awesome improvement!

PR #46, opened 2017-09-08
Fix #42

jabby:
The syntax coloring is ok but I don't understand how language configuration works. 

I add the needed files and the configuration in plugin.xml but it does not work as expected. Maybe @angelozerr can help for this point.

PR #47, opened 2017-09-14
Parent of LESS, SCSS/SASS content type should be CSS

angelozerr:
LESS, SCSS/SASS content types should extends CSS content type. With this change LESS, SCSS/SASS editor could benefit with color feature once LESS, SCSS/SASS editors will be fixed by https://github.com/mickaelistria/eclipse-bluesky/pull/46

### DIFF
--- a/org.eclipse.bluesky/plugin.xml
+++ b/org.eclipse.bluesky/plugin.xml
@@ -87,14 +87,14 @@
    <extension
          point="org.eclipse.core.contenttype.contentTypes">
       <content-type
-            base-type="org.eclipse.bluesky.parent"
+            base-type="org.eclipse.bluesky.css"
             file-extensions="less"
             id="org.eclipse.bluesky.less"
             name="LESS"
             priority="normal">
       </content-type>
       <content-type
-            base-type="org.eclipse.bluesky.parent"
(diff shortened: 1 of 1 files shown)


mickaelistria:
I think that would introduce an issue, because if we make LESS and SCSS content-type extend CSS, then it makes that consumers of a .less file might consider it as a CSS by checking the content-type, pass it to a CSS parser, and the CSS parser would fail. A CSS is a valid LESS or SCSS file, but there is now way to express this with inheritance of content-types in general.
I'd rather avoiding that and have the reconcilier declared on each of the supported content types.

angelozerr:
My idea is that LESS SCSS extends the CSS features like color and after you override features (like language configuration, textmate done by #46)

But if you think it's dangerous, I'm wiating for #46 to do a new PR to support color.

mickaelistria:
yes, I think it's dangerous because someone doing something like `contentTypeManager.getContentType(myFile).contains(cssContentType)` will most likely assume the file is a valid CSS file, and if it's a LESS or SCSS file, it's not a valid CSS file.

angelozerr:
Ok thanks @mickaelistria for your clarification.

PR #52, opened 2017-09-18
Activate color symbol for less + scss.

angelozerr:
Activate color symbol for less + scss.

Signed-off-by: angelozerr <angelo.zerr@gmail.com>


mickaelistria:
Thank you!

PR #80, opened 2018-10-11
Bug 539991 - Try to add the known_hosts

mickaelistria:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,8 @@ pipeline {
 	}
   agent {
     kubernetes {
-      label 'buildtestPod'
+      label 'build-test-pod'
+      defaultContainer 'jnlp'
       yaml """
 apiVersion: v1
 kind: Pod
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=539991

mbarbero:
You should not even need the "defaultMode: 420". I was too quick pasting another config.

mickaelistria:
I've updated the PR, but build is still failing because of known_hosts.
What I don't understand is whether the volume is mounted automatically to the default jnlp container that's used for the copy? Should I explicitly declare a 'jnlp' container in the podTemplate, referencing the eclipse CBI jnlp image directly? What's the name of the docker image so I can reference it?

mbarbero:
Yes, you should add a jnlp image and mount it inside it (or just use the other container to push). That's a limitation from the Kubernetes Jenkins plugin. The docker image is

eclipsecbi/jenkins-jnlp-agent@sha256:356386dcc1fafd296d8dbe206831c73e061ce603bd371810785454c064cefcb1

Issue #95, opened 2018-12-10
tsserver et al cannot be executed


stephan-herrmann:
After installing wildwebdeveloper via Oomph none of the language server functionality worked.
All I got was a log message essentially saying: `[ERROR: spawn EACCES]`. No hint at what was tried to access.

After a lot of aimlessly poking around, I first found, that some "executables" didn't have executable permission (on linux), so I ran `chmod u+x` for them: 
- `vscode-languageserver/bin/installServerIntoExtension`
- `typescript-language-server/lib/cli.js`
- `typescript/bin/tsserver`
- `typescript/bin/tsc`
- `vscode-json-languageserver/bin/vscode-json-languageserver`

Still no luck.

Much later I found the directory `language-servers/node_modules/.bin` and inside I found some broken text files. From comparison with a checked out workspace of wildwebdeveloper I learned, that those files were supposed to be symbolic links, but instead they were just files containing the target file name padded with several trailing null-bytes.

BTW, I couldn't see, what unpacked the wwd jar file, since the feature says `unpack="false"`. Is it actually the `FileLocator.toFileURL()` call in `JSTSLanguageServer` that triggers this unpacking? Anyway, either the build process needs to avoid symbolic links or the unpack routine must restore them as links.

stephan-herrmann:
BTW, I only mentioned Oomph, because that was the only difference between working and not-working installation that I could think of. Take it as a wild guess - might be totally unrelated.

jabby:
Thanks for your report. 
I'm not familiar with Oomph. How did you install a plugin through Oomph?

stephan-herrmann:
Also in that fresh install mentioned above I need to do (at least) these commands to get any typescript support in the IDE:
```
$ cd eclipse/plugins/org.eclipse.wildwebdeveloper_0.1.0.201811161652/language-servers/node_modules/.bin
$ rm tsserver
$ ln -s ../typescript/bin/tsserver
$ rm tsc
$ ln -s ../typescript/bin/tsc
$ chmod u+x tsserver tsc
```

Then restart Eclipse.

Certainly not a procedure that I can expect of those colleagues whom I try to convince to migrate towards Eclipse for their JS/TS development ;p

mickaelistria:
Maybe this works for us because`tsserver` and `tsc` are available in other locations as well and uses them? We definitely need automated tests for Wild Web Developers to spot those issues.

mickaelistria:
[...]

PR #98, opened 2018-12-15
tsserver et al cannot be executed #95

stephan-herrmann:
Scripts for `.bin` as created by `npm` under windows + git bash, after disabling other language servers which cannot be installed on windows.


org.eclipse.wildwebdeveloper/language-servers/.gitignore
```diff
@@ -1,4 +1,5 @@
 /node_modules/
+!/node_modules/.bin
```
mickaelistria: I don't think we should put binaries on SCM herel it's a bad practice and is adding 1 extra set of artifacts to think about maintaining. Those should be provisioned at build time instead.

org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -55,6 +66,7 @@
 							<executable>npm</executable>
 							<arguments>
 								<arg>install</arg>
+								<arg>--no-bin-links</arg>
```
mickaelistria: That's interesting. Maybe there is an option to copy the binaries instead of linking them?

stephan-herrmann:
Giving up on this for now, sorry.

Results: in a previously deployed instance of wildwebdeveloper, several alternatives worked fine:

- repaired symlink
- node script as shown in the patch in the issue
- shell script as generated by npm install on windows (the one that checks if on CYGWIN)

During surefire tests I could get none of these to work. Special difficulty: while I know that cli.js was found & executed I didn't succeed to make any visible logging from js to figure out where exactly it breaks.

Interestingly, npm install on windows does create scripts instead of symlinks, but that build is broken for any of the vscode language servers :(

org.eclipse.wildwebdeveloper/language-servers/.gitignore
stephan-herrmann: > I don't think we should put binaries on SCM herel it's a bad practice and is adding 1 extra set of artifacts to think about maintaining. Those should be provisioned at build time instead.

Feel free to change, but I'm unable to see a simpler alternative (and don't know enough about npm to make it behave "a bit like windows" (i.e., create scripts not links) while running on linux).
I also wouldn't consider scripts as binaries :)

mickaelistria: Right, I used the wrong terms. It's more that I'm reluctant to check-in 3rd party code in general.

mickaelistria:
Further investigation show that best place to improve this is the language server, and that an improvement is already implemented. Replaced this PR by https://github.com/eclipse/wildwebdeveloper/pull/99

PR #99, opened 2018-12-16
Issue #95 - Use newer typescript-language-server

mickaelistria:
Which better resolves tsserver location
+ avoid symlinks in npm


mickaelistria:
@jabby How am I supposed to update the package-lock.json file in that case?

jabby:
Before asking how you should update the package.json. Can you give me the version of npm you use?

mickaelistria:
Locally, I use 7.10 which works fine; but CI uses 8.x which is different (in a breaking way). I'm tweaking the build to work on both versions.

jabby:
I see you get rid of package-lock.json. 
Fine to me for a first step. I'm not sure it's the best idea in future but it implies to use a recent version of npm. I assume you have npm  < 5.x

jabby:
By the way, if you don't want to generate a package-lock.json file on the CI, you should use the following option in npm install: `--no-package-lock`
See https://docs.npmjs.com/cli/install for more explanation.

mickaelistria:
I'm fine committing the package-lock.json, I just don't know how to do it.
It doesn't seem generated locally.

mickaelistria:
Merged as 2ab9d826

jabby:
> I'm fine committing the package-lock.json, I just don't know how to do it. It doesn't seem generated locally.

In fact it is generated with recent version of npm. As npm is package with node.js, you should update to a newer version of npm. 
One other way (I think it the easiest one), you can use this project in order to change easily you node and npm version. https://github.com/creationix/nvm

mickaelistria:
@jabby OK, thanks. So my 7.x version of npm doesn't generate them. If you have a more recent version of npm than I have, feel free to add this package-lock.json file.

PR #102, opened 2018-12-20
Use newer node support from VSCode

mickaelistria:

### DIFF
--- a/org.eclipse.wildwebdeveloper/language-servers/package.json
+++ b/org.eclipse.wildwebdeveloper/language-servers/package.json
@@ -7,11 +7,11 @@
   "license": "EPL",
   "scripts": {},
   "dependencies": {
+    "vscode-css-languageserver": "./vscode-css-languageserver-1.0.0.tgz",
     "vscode-html-languageserver": "./vscode-html-languageserver-1.0.0.tgz",
     "vscode-json-languageserver": "./vscode-json-languageserver-1.0.1.tgz",
-    "vscode-css-languageserver": "./vscode-css-languageserver-1.0.0.tgz",
-    "typescript-language-server": "0.3.7",
     "typescript": "2.8.3",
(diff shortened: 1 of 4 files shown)


mickaelistria:
@jabby: can you please try this PR? You first need a local "mvn clean verify" before you can try it in your IDE.

org.eclipse.wildwebdeveloper/language-servers/package.json
```diff
@@ -7,11 +7,11 @@
...
   "dependencies": {
+    "vscode-css-languageserver": "./vscode-css-languageserver-1.0.0.tgz",
     "vscode-html-languageserver": "./vscode-html-languageserver-1.0.0.tgz",
     "vscode-json-languageserver": "./vscode-json-languageserver-1.0.1.tgz",
-    "vscode-css-languageserver": "./vscode-css-languageserver-1.0.0.tgz",
-    "typescript-language-server": "0.3.7",
     "typescript": "2.8.3",
-    "node-debug": "./node-debug-1.25.4.tgz"
+    "typescript-language-server": "0.3.7",
+    "node-debug2": "./node-debug2-1.25.5.tgz"
```
jabby: I checked on the corresponding github repository. By changing `node-debug` to `node-debug2` we lose the support of all node version previous to 8.0. 
I'm not sure we want explicitly drop this support. In other hand, I think nobody will be impacted as it's quite a new feature in wild web developer.

mickaelistria: Let's just not care about older versions and look ahead. If this patch works for you, feel free to merge it.

jabby: It works for me. By the way I tried with node 8.12.0 and 7.10.1 and both works. 😊

Issue #112, opened 2019-01-08
Wild web developer has useful code that could be used in other places


jonahgraham:
I would like to move a little bit of code higher up the stack from wildwebdeveloper to lsp4e.debug. Specifically InitializeLaunchConfigurations to make it easier for CDT's debug adapter and wildwebdeveloper's adapter to be more consistent on where to get node from.

I see this as a pre-cursor to having Node shipped in some way with CDT/wildwebdev.

The corresponding bug in lsp4e is [Bug 543280](https://bugs.eclipse.org/bugs/show_bug.cgi?id=543280)

mickaelistria:
I think this somehow relates to #23 and detection of node.js should be in a separate bundle and not in main LSP4E bundle.
What do you think about Wild Web Developer hosting those node utils and CDT reusing them?

jonahgraham:
> What do you think about Wild Web Developer hosting those node utils and CDT reusing them?

OK, but only as a separate bundle.

mickaelistria:
> OK, but only as a separate bundle.

Of course!

guw:
@jonahgraham @mickaelistria What is the decision/action items on this one? I have a similar requirement, i.e. I'm about to build an Eclipse plug-in that ships a language server which requires Node at runtime. For best out of the box experience I'm investigating embedding Node in Eclipse.

I like to re-use anything that's been created here.

jonahgraham:
Hi @guw, +1 to what Mickael said. I would love to find the time to work on this, but it is low priority for me too. At the moment my energy is on a special project :-) https://twitter.com/JonahGrahamKC/status/1098531058100944898

mickaelistria:
Related https://github.com/eclipse/wildwebdeveloper/issues/23

mickaelistria:
The node.js detector/embedder is now in place, in its own feature, and documented for external usage: https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper.embedder.node/README.md . The current state IMO covers the initial request, so I'm closing it.
Testing, feedback and improvements are highly welcome (in new issues and new PRs).

jonahgraham:
Thank you @mickaelistria - I will encourage CDT devs to consume that code.

PR #114, opened 2019-01-08
Issue #112: Move InitializeLaunchConfigurations to lsp4e.debug

jonahgraham:
Signed-off-by: Jonah Graham <jonah@kichwacoders.com>


mickaelistria:
@jonahgraham I'm closing as WONTFIX, I hope you don't mind.
I think it's relevant that the support for Node-based language servers remains either in Wild Web Developer, or get extracted to some common place to better handle and share Node stuff; and at the same time, I don't think we should extend the list of duties or goals of LSP4E with node-specific stuff.

Issue #120, opened 2019-01-24
[XML] Support for DTD


mickaelistria:
It seems like XML-LS has support for `.dtd` files.
The `.dtd` should be declared as supported type and linked with the LS.

mickaelistria:
cc @angelozerr Am I right?
@yanxi0830 Something you could look at at some point.

mickaelistria:
Done with #123 . Thanks @yanxi0830 !

PR #123, opened 2019-01-24
Add DTD content type

yanxi0830:
#120 
Signed-off-by: Xi Yan xixiyan@redhat.com

### DIFF
--- a/org.eclipse.wildwebdeveloper/plugin.xml
+++ b/org.eclipse.wildwebdeveloper/plugin.xml
@@ -346,7 +346,7 @@
             priority="normal">
       </content-type>
       <content-type
-            base-type="org.eclipse.wildwebdeveloper.parent"
+            base-type="org.eclipse.wildwebdeveloper.xml"
             file-extensions="xsd"
             id="org.eclipse.wildwebdeveloper.xsd"
             name="XSD"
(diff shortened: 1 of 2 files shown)


org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -345,6 +345,20 @@
             name="XSL"
             priority="normal">
       </content-type>
+      <content-type
+            base-type="org.eclipse.wildwebdeveloper.parent"
```
mickaelistria: Base type for XSD should be the XML type.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -382,6 +404,22 @@
             contentTypeId="org.eclipse.wildwebdeveloper.xsl"
             scopeName="source.xml.xsl">
       </scopeNameContentTypeBinding>
+      <grammar
```
mickaelistria: You don't need to redeclare the grammar as it's already declared above. Just the content-type binding is enough.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -382,6 +404,22 @@
...
+            path="grammars/xml/xml.tmLanguage.json"
+            scopeName="source.xml">
+      </grammar>
+      <scopeNameContentTypeBinding
+            contentTypeId="org.eclipse.wildwebdeveloper.xsd"
+            scopeName="source.xml">
+      </scopeNameContentTypeBinding>
+      <grammar
```
mickaelistria: Same thing.

PR #130, opened 2019-02-11
First support of schemastore

jabby:

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestLanguageServers.java
+++ b/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestLanguageServers.java
@@ -175,7 +175,7 @@ public class TestLanguageServers {
 					return false;
 				}
 			}
-		}.waitForCondition(PlatformUI.getWorkbench().getDisplay(), 3000));
+		}.waitForCondition(PlatformUI.getWorkbench().getDisplay(), 10000));
 	}
 
 	@Test
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/schema/Tuple.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/schema/Tuple.java
@@ -1,0 +1,29 @@
+package org.eclipse.wildwebdeveloper.json.schema;
+
+public class Tuple<T, U> {
+	
+	public T t1;
+	public U t2;
+
+	public Tuple(T t1, U t2) {
+		this.t1 = t1;
+		this.t2 = t2;
+	}
+
+	public T getT1() {
+		return t1;
+	}
+
+	public void setT1(T t1) {
+		this.t1 = t1;
+	}
+
+	public U getT2() {
+		return t2;
+	}
+
+	public void setT2(U t2) {
+		this.t2 = t2;
+	}
+
+}
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSonLanguageServer.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSonLanguageServer.java
@@ -22,8 +22,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import org.eclipse.wildwebdeveloper.Activator;
-import org.eclipse.wildwebdeveloper.InitializeLaunchConfigurations;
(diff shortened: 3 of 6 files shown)


jabby:
Hello @mickaelistria and @angelozerr 
I draft a quick and dirty feature around schemastore. 
I would like to have your first review about my code and feeling about the use of schemastore.org

jabby:
I'm not sure it's a good idea but I increase the timeout of JSON language server test. It takes too much time to parse the catalog of schemastore on the CI.

mickaelistria:
Shouldn't it be an internal feature of the language server?

jabby:
I agree it should be an internal feature of json language server. It's not already developed. There is this issue about the schemastore catalog. https://github.com/Microsoft/vscode/issues/26289
I don't have time to improve vscode-json-languageserver.

This draft is a first step. I am working on a preferences page to manage other schemas. I would like to merge the user defined schemas with schemas from schemastore and send the list to the language server. 
If there is an update on the vscode issue, I will update the code in Wild Web Developer and remove the call to schemastore catalog.

mickaelistria:
Actually, from our Red Hat perspective, we do not really want to invest in IDE-specific support when this would better be in the LS directly and profit to a lot of IDEs at once.
If you want to merge it as some point, feel free; but just beware that it's not what's best and the approach some other contributors would favor and support.

jabby:
`this would better be in the LS directly and profit to a lot of IDEs at once.`

I will check what I can do on this. No promise from my side.

jabby:
For information I created a patch for managing schemastore catalog in vscode-json-languageserver. See https://github.com/Microsoft/vscode/pull/69132 if your are interested.

mickaelistria:
@yanxi0830 Would you like to review this one? It seems related to the preference page you've set up. The goal is that we avoid placing in Wild Web Developer anything that would be better in the language server, but should provide the necessary UI for stuff that users may really need to control by themselves and for which there is no more standard way of doing things (like a config file).

jabby:
FYI information here are the current available options on vscode-json-languageserver https://github.com/Microsoft/vscode/tree/master/extensions/json-language-features/server#settings

yanxi0830:
Since this is a temporary workaround for https://github.com/Microsoft/vscode/issues/26289, would it be better to hardcode the schema associations in the schema catalog like how it was done for some files via `associations.put("package.json", Arrays.asList("http://json.schemastore.org/package"));`? This might avoid the time it takes to parse the catalogs on the client side. 

We could have a preference page that allows users to set custom schemas outside the catalog via `workspace/didChangeConfiguration` settings and an option that enable/disable using catalogs once the ls supports it. WDYT?

mickaelistria:
I like the idea of starting with a 1st iteration hardcoding some schemas (we can even generate those from the schemastore), and then adding the preference page as a 2nd step. Introducing the preference bring hard questions that are not blockers at the moment such as 1. Do we present hardcoded schemas to end-users, 2. Do we let end-user reassociate/override pre-defined schema, 3. How to deal with files that don't have the right extension (need to dynamically create a content-type in user space)?....

org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestLanguageServers.java
```diff
@@ -175,7 +175,7 @@ protected boolean condition() {
 					return false;
 				}
 			}
-		}.waitForCondition(PlatformUI.getWorkbench().getDisplay(), 3000));
+		}.waitForCondition(PlatformUI.getWorkbench().getDisplay(), 10000));
```
mickaelistria: The fact that you need to grow the timeout here shows that pulling the catalog should be happening asynchronously IMHO.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/schema/Tuple.java
```diff
@@ -0,0 +1,29 @@
+package org.eclipse.wildwebdeveloper.json.schema;
+
+public class Tuple<T, U> {
```
mickaelistria: You may prefer to import org.eclipse.xtext.xbase.lib (that's already a transitive dependency) and use its Pair class.

jabby:
@mickaelistria I don't want to change anything in this PR. I think we should close this one and I will focus only on https://github.com/Microsoft/vscode/pull/69132
What do you think about my proposal?

mickaelistria:
I'm totally fine with your proposal, it's much better IMO.
Closed in favor of https://github.com/Microsoft/vscode/pull/69132

Issue #56, opened 2017-09-29
Integrates YAML language server


angelozerr:
As bluesky provides a YAML support, it should be very cool to Integrate YAML Language server https://github.com/redhat-developer/yaml-language-server

mickaelistria:
I think we already have this integration working for Red Hat Developers Studio. I did bootstrap it months ago for Kubernetes: http://www.screencast.com/users/mistria/folders/Default/media/b66f72b7-445d-4675-900a-aab3f484be79 and I think some other developers inside Red Hat have continued to work on it. I'll investigate and try to get this migrated to BlueSky.

mickaelistria:
So this was merged.
I already edited the marketplace entry to associate Wild Web Developer with .yaml and .yml files, so the number of installation should grow naturally.
@yanxi0830 Do you want to take care about adding your video to README and twitting about it (linking to video I guess would be best) ?

PR #132, opened 2019-02-20
Integrate YAML LS

yanxi0830:
#56 
This integrates https://github.com/redhat-developer/yaml-language-server, and adds a preference page to associate schemas.

### DIFF
--- a/org.eclipse.wildwebdeveloper/language-servers/package.json
+++ b/org.eclipse.wildwebdeveloper/language-servers/package.json
@@ -12,6 +12,7 @@
     "vscode-json-languageserver": "./vscode-json-languageserver-1.0.1.tgz",
     "typescript": "2.8.3",
     "typescript-language-server": "0.3.7",
-    "node-debug2": "./node-debug2-1.25.5.tgz"
+    "node-debug2": "./node-debug2-1.25.5.tgz",
+    "yaml-language-server": "0.3.2"
   }
 }
--- a/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestLanguageServers.java
+++ b/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestLanguageServers.java
@@ -17,6 +17,7 @@ import static org.junit.Assert.assertTrue;
 import java.io.ByteArrayInputStream;
 
 import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.ResourcesPlugin;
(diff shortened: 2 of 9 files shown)


mickaelistria:
Why do you think the preference page for schemas is necessary here? It seemed to me that file/schema mapping was implemented directly in the Language Server (so we wouldn't need extra support for it), isn't it @JPinkney ?

org.eclipse.wildwebdeveloper/language-servers/package.json
```diff
@@ -12,6 +12,7 @@
     "vscode-json-languageserver": "./vscode-json-languageserver-1.0.1.tgz",
     "typescript": "2.8.3",
     "typescript-language-server": "0.3.7",
-    "node-debug2": "./node-debug2-1.25.5.tgz"
+    "node-debug2": "./node-debug2-1.25.5.tgz",
+    "yaml-language-server": "0.2.1"
```
mickaelistria: Why not 0.3.2 ?

JPinkney:
@mickaelistria By default a lot of common schemas for YAML files are pulled in from the JSON schema store but if a user wants to set their own schema thats not in the schema store they'd need the preference page

jabby:
@JPinkney I agree with you. The problem will be the same for the json language server. I try to provide a patch to vscode in order to use schemastore catalog and I realized that the  json language server provide a way to add user schema associations. 

Maybe a common preferences for user configuration will be a good idea.

yanxi0830:
the vscode extension allows schema association through its common user/workspace settings, so I think the preference page would be the eclipse equivalent for doing so, unless there is a way to configure it globally like vscode without the preference page?

mickaelistria:
> Maybe a common preferences for user configuration will be a good idea.

At least in Eclipse IDE, if a common preference page for JSon and Yaml schema can work, it can be a good thing to consider.
And if it backs to the same settings of underlying language servers, it'd be nice too.

> I think the preference page would be the eclipse equivalent for doing so, unless there is a way to configure it globally like vscode without the preference page?

Let's stick to a preference page at the moment, but my impression based on how XML schema were deployed is that users should basically never have to deal with a local schema association, since a good schema format is "self-described" (ie for a given file, there can be a flag describing what's the schema to use and where to retrieve it and that's a superior workflow than putting such responsibility in end-users' hands)

org.eclipse.wildwebdeveloper/language-servers/package.json
mickaelistria: @yanxi0830 Your answer here is important as we cannot start the IP approval for yaml-ls before we're sure of which version to take (IP approvals are version specific).

yanxi0830: I was following the version used in https://github.com/redhat-developer/vscode-yaml, which was just updated from 0.2.1 to 0.3.2 a couple of days ago, so 0.3.2 would be the version to take.

mickaelistria:
Here is the request to IP team for inclusion of Yaml LS.

mickaelistria:
@yanxi0830 Would you mind squashing those commits as a single one to clarify history?

yanxi0830:
> > Maybe a common preferences for user configuration will be a good idea.
> 
> At least in Eclipse IDE, if a common preference page for JSon and Yaml schema can work, it can be a good thing to consider.
> And if it backs to the same settings of underlying language servers, it'd be nice too.
> 

I can see that the Json LS and Yaml LS allows for similar settings of schema which can use the same schema-globPattern preference page. 

> there can be a flag describing what's the schema to use and where to retrieve it and that's a superior workflow than putting such responsibility in end-users' hands)

I think that at least with custom yaml files, the yaml ls requires users to sets schema. With this patch, I added the kubernetes schema by default since it breaks the diagnostics test if no schema are set.

PR #146, opened 2019-03-15
Add @angular/language-service plugin to typescript-language-server

yanxi0830:
Instead of adding @angular/language-service to each Angular project, this gives Angular support to typescript files by default. To test: @Component/template in TS files should autocomplete/validate even when plugin is not installed in project. 

@mickaelistria is this the right way to bump version?


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/TypeScriptPlugin.java
```diff
@@ -0,0 +1,21 @@
+package org.eclipse.wildwebdeveloper.jsts;
```
mickaelistria: Need copyright header (copy it from other files in the project)

org.eclipse.wildwebdeveloper.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.2.0.qualifier"
+      version="0.2.1.qualifier"
```
mickaelistria: Since it's a relatively noticeable addition, we'll bump to 0.3.0 to showcase there is more than maintenance fix.

org.eclipse.wildwebdeveloper.feature/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.eclipse.wildwebdeveloper</groupId>
 		<artifactId>parent</artifactId>
-		<version>0.2.0-SNAPSHOT</version>
+		<version>0.2.1-SNAPSHOT</version>
```
mickaelistria: Parent version doesn't have to change since we didn't change any build-related stuff.

org.eclipse.wildwebdeveloper.tests/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for WildWebDeveloper
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper.tests
-Bundle-Version: 0.2.0.qualifier
+Bundle-Version: 0.2.1.qualifier
```
mickaelistria: Since it's a relatively noticeable addition, we'll bump to 0.3.0 to showcase there is more than maintenance fix.

org.eclipse.wildwebdeveloper.xml/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: XML support using Language Server
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper.xml;singleton:=true
-Bundle-Version: 0.1.0.qualifier
+Bundle-Version: 0.1.1.qualifier
```
mickaelistria: The XML bundle wasn't modified, so no need to produce a new version so far. Let's keep current 0.1.0 until we modify something in the bundle.

org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
```diff
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: Wild Web Developer: web development in Eclipse IDE
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper;singleton:=true
 Automatic-Module-Name: org.eclipse.wildwebdeveloper
-Bundle-Version: 0.2.0.qualifier
+Bundle-Version: 0.2.1.qualifier
```
mickaelistria: Since it's a relatively noticeable addition, we'll bump to 0.3.0 to showcase there is more than maintenance fix.

org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -1,15 +1,13 @@
...
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 	<artifactId>org.eclipse.wildwebdeveloper</artifactId>
 	<parent>
 		<groupId>org.eclipse.wildwebdeveloper</groupId>
 		<artifactId>parent</artifactId>
-		<version>0.2.0-SNAPSHOT</version>
+		<version>0.2.1-SNAPSHOT</version>
```
mickaelistria: Since it's a relatively noticeable addition, we'll bump to 0.3.0 to showcase there is more than maintenance fix.

pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.eclipse.wildwebdeveloper</groupId>
   <artifactId>parent</artifactId>
-  <version>0.2.0-SNAPSHOT</version>
+  <version>0.2.1-SNAPSHOT</version>
```
mickaelistria: No need to change parent version if no build-related (the parent or the target platform) stuff did change.

pom.xml
```diff
@@ -43,7 +43,7 @@
 						<artifact>
 							<groupId>org.eclipse.wildwebdeveloper</groupId>
 							<artifactId>target-platform</artifactId>
-							<version>0.1.0-SNAPSHOT</version>
+							<version>0.1.1-SNAPSHOT</version>
```
mickaelistria: No need to change TP version if it didn't change.

repository/category.xml
```diff
@@ -1,15 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <site>
-   <feature id="org.eclipse.wildwebdeveloper.feature" version="0.2.0.qualifier">
+   <feature id="org.eclipse.wildwebdeveloper.feature" version="0.2.1.qualifier">
```
mickaelistria: The simplest is to put `version="0.0.0"` and it will be dynamically set.

repository/pom.xml
```diff
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>org.eclipse.wildwebdeveloper</groupId>
 		<artifactId>parent</artifactId>
-		<version>0.2.0-SNAPSHOT</version>
+		<version>0.2.1-SNAPSHOT</version>
```
mickaelistria: Since it's a relatively noticeable addition, we'll bump to 0.3.0 to showcase there is more than maintenance fix.

target-platform/pom.xml
```diff
@@ -4,8 +4,8 @@
 	<parent>
 		<groupId>org.eclipse.wildwebdeveloper</groupId>
 		<artifactId>parent</artifactId>
-		<version>0.2.0-SNAPSHOT</version>
+		<version>0.2.1-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-target-definition</packaging>
-	<version>0.1.0-SNAPSHOT</version>
+	<version>0.1.1-SNAPSHOT</version>
```
mickaelistria: Since there was no change, no need to change the version.

angelozerr:
At first thank's @yanxi0830 for your PR!

I wanted just to give you some feedback about Angular support in https://github.com/angelozerr/angular-eclipse about version of angular language service. I think it's important to give the capability to update the angular language service by user (instead of waiting for new version of wwd).

mickaelistria:
> I think it's important to give the capability to update the angular language service by user (instead of waiting for new version of wwd).

We should cover this story in another patch.
We can release Wild Web Developer pretty often, so I don't think it's worth worrying too much about it right now.

mickaelistria:
CQ https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19296

mickaelistria:
CQ approved.

yanxi0830:
> I think it's important to give the capability to update the angular language service by user (instead of waiting for new version of wwd).

Can the users update the angular language service in their own angular project with package.json and tsconfig.json, which is provided by default by tsserver?

org.eclipse.wildwebdeveloper.tests/pom.xml
```diff
@@ -35,6 +35,29 @@
 					<useUIHarness>true</useUIHarness>
 				</configuration>
 			</plugin>
+			<plugin>
```
mickaelistria: If you commit the angular-app (which is a good thing IMO), you don't need to put the logic to dynamically fetch them at build-time in the pom.xml, do you?

Or you can remove the content files and keep this block instead.

If you decide to keep the sources checked-in (that'd be my favorite), can you please trim the application to the very minimal set of files and contents for the unit test to be relevant?

yanxi0830: This is still necessary to perform an npm install for angular dependencies, otherwise the tsserver will report import errors rather than template errors, which doesn't tell if the angular service is working or not. And I'll check to see if any sources/dependencies can be removed.

mickaelistria: Ok.
Another possibility would be to perform the `npm` step during test execution. It makes things a bit simpler to maintain, but add some extra time to test every time.

yanxi0830: Sure, updated the patch to do npm install at test time.

org.eclipse.wildwebdeveloper.xml/META-INF/MANIFEST.MF
mickaelistria: As it's not changed, this needs to stay to 0.1.0

org.eclipse.wildwebdeveloper.xml/pom.xml
```diff
@@ -9,7 +7,7 @@
 		<version>0.2.0-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-plugin</packaging>
-	<version>0.1.0-SNAPSHOT</version>
+	<version>0.1.1-SNAPSHOT</version>
```
mickaelistria: As it's not changed, this needs to stay to 0.1.0

org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestLanguageServers.java
```diff
@@ -233,5 +248,70 @@ protected boolean condition() {
...
+		File folder = new File(url.getFile());
+		if (folder.exists()) {
+			FileUtils.copyDirectory(folder, project.getLocation().toFile());
+			Process process = new ProcessBuilder(getNpmLocation(), "install", "--no-bin-links", "--ignore-scripts").directory(project.getLocation().toFile()).start();
+			while (process.isAlive());
+			project.refreshLocal(IResource.DEPTH_INFINITE, new NullProgressMonitor());			
+			IFile file = project.getFolder("src").getFolder("app").getFile("app.component.ts");
+			String fileContent = IOUtils.toString(file.getContents(), StandardCharsets.UTF_8);
+			ITextEditor editor = (ITextEditor) IDE.openEditor(PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage(), file);
+			editor.getDocumentProvider().getDocument(editor.getEditorInput()).set(fileContent);
```
mickaelistria: Does that line change anything? It's setting document with same contents as previously. Is it "only" to trigger an event?
If so, maybe it'd be best if you start from a clean file, check there are not markers and then make a modification on the document and see that markers are added accordingly.

org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestLanguageServers.java
```diff
@@ -233,5 +248,70 @@ protected boolean condition() {
...
+					}
+				}
+			}.waitForCondition(PlatformUI.getWorkbench().getDisplay(), 50000));
+			
+			IMarker [] markers = file.findMarkers("org.eclipse.lsp4e.diagnostic", true, IResource.DEPTH_ZERO);
+			boolean foundError = false;
+			for (IMarker marker : markers) {
+				int lineNumber = marker.getAttribute(IMarker.LINE_NUMBER, -1);
+				if (lineNumber == 6) {
+					foundError = true;
```
mickaelistria: Would be good to check also for a keyword like "Component" in the marker message to ensure the message comes from the right LS.

org.eclipse.wildwebdeveloper.tests/testProjects/angular-app/.gitignore
```diff
@@ -0,0 +1,46 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
```
mickaelistria: I think you can remove this file.

org.eclipse.wildwebdeveloper.tests/testProjects/angular-app/angular.json
```diff
@@ -0,0 +1,136 @@
...
+                {
+                  "type": "initial",
+                  "maximumWarning": "2mb",
+                  "maximumError": "5mb"
+                }
+              ]
+            }
+          }
+        },
+        "serve": {
```
mickaelistria: I think you can remove this goals and next ones and keep only `build` (you may even be able to remove build)

org.eclipse.wildwebdeveloper.tests/testProjects/angular-app/angular.json
```diff
@@ -0,0 +1,136 @@
...
+              "src/tsconfig.spec.json"
+            ],
+            "exclude": [
+              "**/node_modules/**"
+            ]
+          }
+        }
+      }
+    },
+    "angular-app-e2e": {
```
mickaelistria: You can probably fully remove this `project`

org.eclipse.wildwebdeveloper.tests/testProjects/angular-app/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@angular/forms": "~7.2.0",
+    "@angular/platform-browser": "~7.2.0",
+    "@angular/platform-browser-dynamic": "~7.2.0",
+    "@angular/router": "~7.2.0",
+    "core-js": "^2.5.4",
+    "rxjs": "~6.3.3",
+    "tslib": "^1.9.0",
+    "zone.js": "~0.8.26"
+  },
+  "devDependencies": {
```
mickaelistria: You may be able to remove all the devDependencies.

org.eclipse.wildwebdeveloper.tests/testProjects/angular-app/src/app/app.component.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { TestBed, async } from '@angular/core/testing';
```
mickaelistria: You may be able to remove this test file.

org.eclipse.wildwebdeveloper.tests/testProjects/angular-app/tslint.json
```diff
@@ -0,0 +1,75 @@
+{
```
mickaelistria: I think you can remove that file.

mickaelistria:
Thanks @yanxi0830 .
This will stay in Snapshots for some time. We'll make a new release of Wild Web Developer in ~2 weeks that will include this feature, and updates to the various language servers.

PR #166, opened 2019-04-25
Add extended JSON schema support via extension point

ndoschek:
This adds an extended JSON schema support via an extension point. The schema can be declared either by URL or a JSON file.
(as mentioned in PR https://github.com/eclipse/wildwebdeveloper/pull/27)

Remarks:
- Developing a plug-in that uses this extension point:
It is possible to use a JSON schema file, that is stored in another plug-in by declaring an additional bundle-id in the schema association. At the moment, it is necessary to unpack the bundle after installation to access the file at runtime. To achieve this, set the header Eclipse-BundleShape to the value dir.
If it becomes necessary to retrieve the file and keep the bundle as a jar, the implementation of the file locating needs to be adjusted and support a bundle directory.

- There existed a schema association for 'typing.json', but the JSON schema is not available at the JSON schema store anymore, therefore i did not create a schema association for this file pattern.


mickaelistria:
I'm reluctant to add this kind of features at the moment for the following reasons:
1. This makes Wild Web Developer progressively becoming an API, providing extension point and exporting packages while I don't think the project currently have the community to support the huge cost associated with exposing its internals to the outer world and taking care of backward compatibility. Maybe you or your employer can commit to the associated maintenance?
2. This puts in Wild Web Developer some responsibility that is already better covered by the schema store. The Schema Store is external, frequently updated and benefits to the whole domain (more than Wild Web Developer only). I think it's providing more value to Wild Web Developer and the whole dev tools domain to try to enrich the schemastore rather than setting up Wild Web Developer-specific workarounds. Having an extension point may not drive end-users to the better solution that is to extend the Schema Store instead.

So the main question I have is: why don't you put the necessary schemas in the schema store directly?

Then, about implementation, we mustn't force users to use an `Eclipse-BundleShape: dir`. Using something like `FileLocator.toFileURL()` can convert internal URLs to file ones and can allow to work with files inside jarred bundles.

ndoschek:
Thank you for your answer and I apologize for my late reply.
We developed this feature as part of a custom JSON editor prototype and thought about contributing it to your project right away before any decision regarding the final solution was made.
In our case the schemas might be proprietary, so a direct upload to the schema store is therefore not an option unfortunately.

org.eclipse.wildwebdeveloper/schema/org.eclipse.wildwebdeveloper.jsonschemaassociations.exsd
```diff
@@ -0,0 +1,238 @@
...
+      <complexType>
+         <sequence minOccurs="1" maxOccurs="unbounded">
+            <sequence minOccurs="1" maxOccurs="unbounded">
+               <choice>
+                  <element ref="bundleFilePath"/>
+                  <element ref="url"/>
+               </choice>
+            </sequence>
+            <sequence minOccurs="1" maxOccurs="unbounded">
+               <element ref="filePattern"/>
```
mickaelistria: Like everywhere else in Generic Editor and LSP4E, we should use content-types instead of direct file patterns. Content-types provide ability to match patterns and are reusable.

org.eclipse.wildwebdeveloper/schema/org.eclipse.wildwebdeveloper.jsonschemaassociations.exsd
```diff
@@ -0,0 +1,238 @@
...
+   </element>
+
+   <element name="schemaAssociations">
+      <annotation>
+         <documentation>
+            Sequence of all JSON schema associations
+         </documentation>
+      </annotation>
+      <complexType>
+         <sequence minOccurs="1" maxOccurs="unbounded">
```
mickaelistria: I'd rather see extensions of the form

```xml
  <extension point="org.eclipse.wildwebdeveloper.schemaAssociations">
     <schemaAssocitation
          contentTypeId="org.eclipse.wildwebdeveloper.package.json"
          url="http://json.schemastore.org/tsconfig"
          localPath="jsonSchema/tsconfig.schema.json"/>
  </extension>
```

The local path may be used as a fall back if URL is incorrect, so we could allow both and explicit URL has higher priority.

org.eclipse.wildwebdeveloper/schema/org.eclipse.wildwebdeveloper.jsonschemaassociations.exsd
```diff
@@ -0,0 +1,238 @@
...
+            <annotation>
+               <documentation>
+                  The relative path of the JSON schema file within the bundle (i.e. the plug-in).
+               </documentation>
+            </annotation>
+         </attribute>
+         <attribute name="bundleId" type="string">
+            <annotation>
+               <documentation>
+                  An optional symbolic ID of the bundle (i.e. the plug-in ID).
```
mickaelistria: Do you have a use-case for it? If not, you could remove this field.
If you have a use case for it, please consider merging it with the `relativePath` entry, so if it's a path, it will load a resource in the bundle; but it can also allow resource URLs like `platform:/plugin/some.other.bundle/path/to/file.schema.json`. I just found https://lmap.blogspot.com/2008/03/platform-scheme-uri.html which gives similar example for icon loading.

org.eclipse.wildwebdeveloper/schema/org.eclipse.wildwebdeveloper.jsonschemaassociations.exsd
```diff
@@ -0,0 +1,238 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="url">
+      <complexType>
+         <attribute name="url" type="string" use="required">
+            <annotation>
+               <documentation>
+                  JSON schema file location, such as &lt;a href=&quot;http://json.schemastore.org/tsconfig&quot;&gt;http://json.schemastore.org/tsconfig&lt;/a&gt;.
```
mickaelistria: Please explain here what this URL will be used for, something like `this resource indicated by this URL will be fetched and its content will be used as schema. If the fetch operation fails and a <code>relativePath</code> is specified, the file indicated by relativePath will be used as failback.`

org.eclipse.wildwebdeveloper/schema/org.eclipse.wildwebdeveloper.jsonschemaassociations.exsd
```diff
@@ -0,0 +1,238 @@
...
+               </documentation>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="filePattern">
```
mickaelistria: Pattern matching for files is one of the roles of content-type, we don't want to duplicate this responsibility.

org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
```diff
@@ -28,4 +28,6 @@ Require-Bundle: org.eclipse.ui,
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Bundle-ActivationPolicy: lazy
 Eclipse-BundleShape: dir
-Export-Package: org.eclipse.wildwebdeveloper.debug;x-internal:=true
+Export-Package: org.eclipse.wildwebdeveloper.debug;x-internal:=true,
+ org.eclipse.wildwebdeveloper,
```
mickaelistria: Please don't export more packages than necessary.

target-platform/target-platform.target
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?><?pde version="3.8"?>
-<target name="wildwebdeveloper" sequenceNumber="10">
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><?pde version="3.8"?><target name="wildwebdeveloper" sequenceNumber="10">
```
mickaelistria: Useless change, please try to avoid it to reduce noise and make discussions easier to focus on actual payload.

mickaelistria:
Also, please sign the Eclipse Contributor Agreement which is required to get code included in Eclipse projects: https://accounts.eclipse.org/users/ndoschek/eca

mickaelistria:
Also, if you look at the YAML language settings, you can see that addition of schema and file-association is handled successfully via preferences. Maybe it would be more valuable to implement similar preference management for JSON (so user could tweak it) and have your downstream code configuring the preference.

ndoschek:
Thank you for your hint to implement the JSON schema associations via preference pages.
I reverted the extension point implementation and added a new preference page to associate JSON schemas to file patterns. The user can now define the schema location either by adding a valid URL or a local file path.

I added the existing schema associations as default entries (except for the schema 'typing.json', as mentioned in my initial PR comment).

I am looking forward to hear your opinion on this change.
Best regards!

ndoschek:
I open a new PR with the new implementation.

PR #170, opened 2019-06-05
Update to Angular language service 8.0.0

mickaelistria:

### DIFF
--- a/org.eclipse.wildwebdeveloper/language-servers/package-lock.json
+++ b/org.eclipse.wildwebdeveloper/language-servers/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@angular/language-service": {
-      "version": "7.2.9",
-      "resolved": "https://registry.npmjs.org/@angular/language-service/-/language-service-7.2.9.tgz",
-      "integrity": "sha512-+8P3yGqrCZy8PaoxUUlC6cowxPAIbuW4HwhvHvYV3G99LDQUC/bxUPSPhqbrAvDiFr3VeXqivzXgFCTo69S1WQ=="
+      "version": "8.0.0",
+      "resolved": "https://registry.npmjs.org/@angular/language-service/-/language-service-8.0.0.tgz",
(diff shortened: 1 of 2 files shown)


jabby:
Hi @mickaelistria when you update the version of a dependency, you should use the npm cli. 
In fact the cli will also update package-lock.json file for you. 

I suggest to use the same version of npm (install the same version of node, it should be ok).

In case the update is **not** a `major` update, you can use the following command line: `npm update @angular/language-service`. It will update `@angular/language-service` from `7.2.9` to `7.2.15`. 

In **your case** you want to update  `@angular/language-service` to `8.0.0`. To do so, just run `npm install @angular/language-service@8.0.0`.

By the way, I run a quick report on package.json. It seems we have more than one old dependency. 

```
$ npm outdated
Package                        Current  Wanted  Latest  Location
@angular/language-service        7.2.9  7.2.15   8.0.0  wildwebdeveloper-node-languageservers
typescript                       2.8.3   2.8.3   3.5.1  wildwebdeveloper-node-languageservers
typescript-language-server       0.3.7   0.3.7   0.3.8  wildwebdeveloper-node-languageservers
typescript-plugin-css-modules    1.0.5   1.0.5   1.1.1  wildwebdeveloper-node-languageservers
yaml-language-server             0.3.2   0.3.2   0.4.1  wildwebdeveloper-node-languageservers
```

I think we should upgrade all our dependencies. I don't just remember why we don't use the latest version of vscode for some language servers.

mickaelistria:
We cannot and shouldn't update everything at once for a few reasons:
* We need IP approval for each dependency set and those tend to come sequentially
* I tried updating multiple things at once and it breaks a lot of tests and is hard to perform efficiently. Keeping a narrower scope allows more incremental and easier progress.

mickaelistria:
We also see https://github.com/theia-ide/typescript-language-server/issues/109 that prevents us from upgrading all at once (and basically from upgrading most things).

jabby:
Sorry @mickaelistria I was not clear. I know we should not update all dependencies in one commit / PR.

mickaelistria:
Replaced by #188

PR #173, opened 2019-06-27
Update target platform to 4.12 / 2019-06 release

jabby:


mickaelistria:
At this point, as we won't produce a new build for 2019-06, it might be better to go directly to 4.13/2019-09

akurtakov:
It would be nice if we can make extra release targeting 2019-06 with updated lsp4xml.

mickaelistria:
Ok, then let's accept this patch as it.
Thanks @jabby

jabby:
@mickaelistria you're welcome

@akurtakov About lsp4xml. Do you know when lsp4xml 0.8.0 will be released? I will be happy to do the corresponding CQ and eventually merge your PR or create a new one.

PR #175, opened 2019-06-28
Bump version to 0.4.1.

akurtakov:
Opens the stream for development.


mickaelistria:
Wild Web Developer attempts to use SemVer so we try to only bump versions when we make changes.

akurtakov:
The patch doesn't change bundle versions it changes the versions of the "infra" so on the first change this shouldn't happen together with the patch.

mickaelistria:
OK, my bad, I'll merge it when CI says it's fine.

Issue #187, opened 2019-07-10
TypeScript 3.x support


akurtakov:
Support for TypeScript 3.x is missing

akurtakov:
I'll file CQs shortly.

akurtakov:
CQ for typescript https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20338

jabby:
Thanks.

akurtakov:
CQ for typescript-language-server https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20339

akurtakov:
> CQ for typescript-language-server https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20339

APPROVED now.

mickaelistria:
Did you try upgrading then?
The combination of node_modules is relatively fragile, and upgrading is getting more and more complex. Fortunately the tests are there to detect issues early.

akurtakov:
Upgrade typescript? Yes, see the attached PR.

akurtakov:
> CQ for typescript https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20338

APPROVED too

akurtakov:
Adding angulara/language-service to the mix as version 8.0.0 added support for typescript 3.x. CQ is https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19296

akurtakov:
Sigh CQ is https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20410

akurtakov:
> Sigh CQ is https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20410

Approved.

akurtakov:
CQ for typescript-plugin-css-modules 1.2.1  https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20423

akurtakov:
> CQ for typescript-plugin-css-modules 1.2.1 https://dev.eclipse.org/ipzilla/show_bug.cgi?id=20423

Approved now.

PR #188, opened 2019-07-10
Update to typescript 3.5.3 and ls 0.3.8.

akurtakov:
Fix for #187


mickaelistria:
Can you please check locally whether manually reproducing the test case scenario that are failing works as expected or fails too?
The combination of all those LS is getting more and more fragile, and it could be that updating only typescript can cause issue on other parts now.

akurtakov:
All tests pass locally sometimes - TestTypeScript.testHTMLinTSXFile fails randomly but it has been failing that way prior to this patch too.

jabby:
@mickaelistria I agree with you when you said that updating some language servers will broke others. 

I just checked how vscode plugins like vetur (for vue.js) worked. In fact each plugins seems to have his own node_modules. It's duplicate a lot of files between some plugins but updating one does not break others.

May be we should start to separate some language servers into separate plugins? I think we can keep JSON, HTML and CSS language server together as there are part of vscode by default.

mickaelistria:
@jabby Do you know if it's possible to get multiple versions of the same module in node_modules and to have node picking some dependencies that seem more safely compatible?

mickaelistria:
@akurtakov tests should be a bit improved. Please try rebasing and cross fingers ;)

PR #198, opened 2019-07-18
Try to make tests more stable

mickaelistria:
The typescript-language-server sometimes need some time just after
`didOpen` before reacting properly to further request. Add such delay in
tests hoping it's enough to fix the issue.


mickaelistria:
@akurtakov FYI

PR #201, opened 2019-07-18
Make DebugTest more stable

mickaelistria:
Refine conditions to make tests more robust.


mickaelistria:
This was most likely related to the Docker image using an incorrect node.js.
When upgrading the Docker image, I also improved the test and it seem stable now. No more action required AFAIK.

PR #205, opened 2019-07-22
Use vscode-json-languageserver from npmjs

mickaelistria:
Instead of extracting it from VSCode binary

Signed-off-by: Mickael Istria <mistria@redhat.com>

PR #207, opened 2019-07-26
Use npm ci instead of npm install

jabby:

### DIFF
--- a/org.eclipse.wildwebdeveloper/pom.xml
+++ b/org.eclipse.wildwebdeveloper/pom.xml
@@ -79,7 +79,7 @@
 						<configuration>
 							<executable>npm</executable>
 							<arguments>
-								<arg>install</arg>
+								<arg>ci</arg>
 								<arg>--no-bin-links</arg>
 								<arg>--ignore-scripts</arg>
 							</arguments>


mickaelistria:
Thanks!

mickaelistria:
@jabby: it seems like `npm ci --no-bin-links --ignore-scripts` ignore the 2 flags, and using it usually creates the symbolic links, that we don't want to use in Wild Web Developer. Do you know a possible way to use both `npm ci` and avoid creation of symlinks?

jabby:
If I understand correctly the npm documentation (not the best one), it should not infore the 2 flags. https://docs.npmjs.com/cli/ci.html

Maybe problem comes from package-lock.json. I checked and it contains direct local links to tar.gz files.

jabby:
Which symbolic links did you find? I tried on my laptop and at least --ignore-scripts is ok.

If I use `npm ci` I get an error about postinstall script from node-debug2 1.25.5
If I use `npm ci --ignore-scripts`, there is no error. 

And I was not able to find a symbolic link in node_modules.

mickaelistria:
I see symlinks after `npm ci --no-bin-links --ignore-scripts` in the `node_modules/.bin` folder. Those symlinks generate an error later during tests provisioning.

jabby:
FYI 
https://github.com/eclipse/wildwebdeveloper/pull/205#issuecomment-516337617

Issue #220, opened 2019-08-21
Support node.js 12


rombert:
I am using the "current" version of Node.js, which is is 12. I keep getting popups (one per file type I guess, it would be nice to get just one per session at least ) saying that my version of node is not supported, although things seem to mostly work.

![image](https://user-images.githubusercontent.com/21237/63414829-e739a380-c3fc-11e9-9da8-f5c5a7a55ce2.png)

It would be great to fully support Node.js fully.

mickaelistria:
What is it you call current? Latest release or latest snapshot.
@jabby isn't this alredy fixed?

rombert:
@mickaelistria  - I mean Node.js current - https://nodejs.org/en/download/current/ . On the Eclipse side I am using Wild Web Developer - Web edition in Eclipse IDE (Experimental) 0.4.1.201907221412 .

mickaelistria:
Can you please try latest snapshot build at http://download.eclipse.org/wildwebdeveloper/snapshots/ ?

jabby:
@mickaelistria I did not update this. I will provide a patch in few minutes.

jabby:
I created PR #221
I think there is a misunderstanding between the message and what version of node the developer can use for nide projects. 

@rombert If I understand correctly, the message is about the version of node for your project. Am I right? 
To be clear, it's version of node used by Wild Web Developer to enable features like completion / diagnostics / ...

rombert:
Thanks @jabby . To be honest I am not sure about what version of node it is talking about :-) I did not use debugging, launching, etc. I mostly used it for editing SASS and HTML files.

mickaelistria:
Patch merged, issue should be fixed by using upcoming snapshot build.
Thanks @jabby !

rbr9292:
Good morning, sorry to reopen this issue but I'm having the same problem... Already tested node v10 and v12 and still having the same popup... I'm using KDE Neon, latest stable version, and Eclipse latest version as well. 

Thanks in advance,
Renan

![popup](https://user-images.githubusercontent.com/47678800/78889539-38ed7780-7a3a-11ea-8cc2-97d68378f5dc.png)

![node](https://user-images.githubusercontent.com/47678800/78889440-06dc1580-7a3a-11ea-9f35-49df5cf00ba6.png)

mickaelistria:
Please open a new ticket for your case with more details about your OS, how you installed `node`, and what `node --version` returns

[...]

PR #221, opened 2019-08-21
Issue #220 - Enable support for node 11 and 12

jabby:

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
@@ -35,7 +35,7 @@ import org.osgi.framework.Version;
 public class InitializeLaunchConfigurations {
 
 	private static final Set<Integer> SUPPORT_NODEJS_MAJOR_VERSIONS = Collections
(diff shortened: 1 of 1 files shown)


mickaelistria:
Should we instead use a versionRange and assume compatibility with future versions as well?

jabby:
Fair question. It will reduce time to maintain this part of code. Next version of node will be released the 2019-10-22. 
I think it's easiest to react if we find a problem with a new version. In the other hand, I checked in node_modules. Some restrictions suggest, we should not use node in version 7 or 9 or 11. I think it's because these versions are not LTS. 


One example from  `vscode-jsonrpc@4.0.0`
Engine versions are 
```json
{
  ...
  "engines": {
    "node": ">=8.0.0 || >=10.0.0"
  },
  ...
}
```

mickaelistria:
I think we at least need to have support for 13 if it's to be released soon. Some early-adopters will try it and showing them an error message is less interesting to them and to Wild Web Developer than having them using hte tool, and maybe finding issues.

jabby:
I added the version 13 to the list of supported node runtime. 
If you are ok with this change, we can merge. Update 2 times per year this is not so difficult.

mickaelistria:
> If you are ok with this change, we can merge.

Yes, it's a valuable improvement. Feel free to merge it.

> Update 2 times per year this is not so difficult.

I'm more concerned about us remembering to do it at the right time ;)

rombert:
Great, thanks to both of you for the rapid fix.

PR #222, opened 2019-08-21
[WIP] Remove package.json

mickaelistria:
Directly list necessary npm modules in pom.xml for easiler maintenance.


mickaelistria:
At the moment, this makes node debugger fail with

```
Error processing "launch": TypeError: Cannot read property 'getGeneratedPathFromAuthoredPath' of undefined
    at NodeDebugAdapter.<anonymous> (/home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/language-servers/node_modules/node-debug2/out/src/nodeDebugAdapter.js:509:72)
    at Generator.next (<anonymous>)
    at /home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/language-servers/node_modules/node-debug2/out/src/nodeDebugAdapter.js:10:71
    at new Promise (<anonymous>)
    at __awaiter (/home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/language-servers/node_modules/node-debug2/out/src/nodeDebugAdapter.js:6:12)
    at NodeDebugAdapter.resolveProgramPath (/home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/language-servers/node_modules/node-debug2/out/src/nodeDebugAdapter.js:496:16)
    at NodeDebugAdapter.<anonymous> (/home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/language-servers/node_modules/node-debug2/out/src/nodeDebugAdapter.js:128:52)
    at Generator.next (<anonymous>)
    at fulfilled (/home/mistria/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/language-servers/node_modules/node-debug2/out/src/nodeDebugAdapter.js:7:58)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

jabby:
@mickaelistria It's a really bad idea to remove package-lock.json. 
It's the only file which ensure that two differents builds will get the same transitive dependencies. 

For an explanation, if a transitive dependency is not fixed in one package.json of a dependency. We can upload two different versions of the transitive dependency between two builds.

mickaelistria:
@jabby I think it's just better if we get rid of package-lock.json, for the following reasons
1. We don't really care about transitive dependency management, it's a black-box to Wild Web Developer. We only care about things working; and the tests are what tell whether things work or not.
2. We have strategies at build time to reject builds that produce different content for same version, and enforce a version bump in those cases, so even if we're affected by a transitive dependency change, the build will detect it and let us investigate whether to allow it or not
3. We need to be more agile at upgrading versions of the language-servers, and experiences over the last months have shown that upgrading those package*.json file is not robust, very error-prone, hard to grok; and that the process of upgrading language-servers involves multiple file and extra complexity. Getting rid of the package*.json removes a lot of the complexity of steps to perform, and having everything in pom.xml makes it easier for a newcomer to just update stuff without figuring out a complex process.

jabby:
>     1. We don't really care about transitive dependency management, it's a black-box to Wild Web Developer. We only care about things working; and the tests are what tell whether things work or not.

From development point of view, I agree. But from the perspective of licence compliance,  I think it's better to freeze transitive dependencies and only update them when needed.
 
>     2. We have strategies at build time to reject builds that produce different content for same version, and enforce a version bump in those cases, so even if we're affected by a transitive dependency change, the build will detect it and let us investigate whether to allow it or not

I agree we have a great tool to detect if there is a dependency change. But there is a case where the removal of package*.json seems odd to me. 

1. We build a version without package*.json in source code. It will download specific versions of each node dependencies. (transitive or not)
2. Then one transitive dependency which is not frozen has a new minor / fix version. No new build so we are ok.
3. We create a new PR which implement a new feature or a fix in Java code only. We will build again Wild Web Developer. Because it's a fix or a new feature, we will update our version number. Am I right? How the tool will warn us that something change in our dependencies?

More over, why a contribution in Java code like adding a new node runtime should modify the language server part of WWD. 

One other problem I saw is about the lose of functionnalities. If we have an update of our transitive dependency for our servers, we can introduce a regression or a bug. Even if we don't change anything about our main dependencies.

>     3. We need to be more agile at upgrading versions of the language-servers, and experiences over the last months have shown that upgrading those package*.json file is not robust, very error-prone, hard to grok; and that the process of upgrading language-servers involves multiple file and extra complexity. Getting rid of the package*.json removes a lot of the complexity of steps to perform, and having everything in pom.xml makes it easier for a newcomer to just update stuff without figuring out a complex process.

I agree with a big part of your analyse, especially `very error-prone, hard to grok; and that the process of upgrading language-servers involves multiple file and extra complexity`. But trust me, an error because of a transtive dependency is more difficult to fix. 

`having everything in pom.xml makes it easier for a newcomer`: I disagree with this part. It's easier when you don't know the JavaScript dependency world / hell. But for my point of view, it's quite more difficult to understand that node dependencies managed by npm are in fact managed in pom.xml file.

mickaelistria:
Thanks for the discussion @jabby . However, my experience and @akurtakov 's one on that matter have shown that with package.json, we've faced a lot of bugs about transitive dependencies, and that is what has basically prevented from upgrading VSCode for many months. With the suggested approach, upgrading becomes trivial.
We also expect contributors to Wild Web Developer to be mostly Eclipse plugins developers, more aware of Java than they are with JavaScript, and more familiar with pom.xml than they are with package.json and its mechanics (looking at history of contributors, all of them are Eclipse plugin developers).
I think we'll deal with issue of transitive dependencies when we need them.

But your concerns are bringing me to an interesting addition to that change: although we won't commit a package-lock.json, we'll include in in the binary build, so if we have a known working version and see an issue with a current build, we'll be able to diff the package-lock.json of the working version (in the .jar) with the one generated by the current build.
I'll submit a new patch shortly adding that.

jabby:
If @akurtakov and you are ok with this change, I have no problem with the merge. I will update my PR about vuejs tomorrow. I will have time to finish.

mickaelistria:
I just merged it actually.
This change has allowed us to successfully migrate to newer VSCode quite easily compared to previous unsuccessful attempts. I guess you'll also enjoy it for you vuejs work ;)

PR #250, opened 2019-09-05
Java projects shouldn't be able to launch with Firefox #248

### LINKED ISSUE #248: [Firefox Debug] Java projects shouldn't be able to launch with Firefox
With a java project opened, right-clicking on the editor or project and selecting `Debug as...` gives Firefox as a launch option. This doesn't make sense for java projects and must be removed.  
image

AObuchow:
fix  #248

### DIFF
--- a/org.eclipse.wildwebdeveloper/plugin.xml
+++ b/org.eclipse.wildwebdeveloper/plugin.xml
@@ -545,11 +545,25 @@
                   <iterate>
                      <adapt
                            type="org.eclipse.core.resources.IResource">
+                        <test
+                              forcePluginActivation="true"
+                              property="org.eclipse.wildwebdeveloper.isHTMLLaunchable">
+                        </test>
                      </adapt>
                   </iterate>
                </with></enablement>
          </contextualLaunch>
       </shortcut>
+   </extension>
+    <extension
+         point="org.eclipse.core.runtime.adapters">
+      <factory
+            adaptableType="org.eclipse.core.resources.IResource"
+            class="org.eclipse.wildwebdeveloper.debug.HTMLLaunchableAdapterFactory">
+         <adapter
+               type="org.eclipse.debug.ui.actions.ILaunchable">
+         </adapter>
+      </factory>
    </extension>
    <extension
          point="org.eclipse.core.expressions.propertyTesters">
@@ -558,6 +572,13 @@
             id="org.eclipse.wildwebdeveloper.propertyTester.isNodeResource"
             namespace="org.eclipse.wildwebdeveloper"
             properties="isNodeResource"
+            type="java.lang.Object">
+      </propertyTester>
+      <propertyTester
+            class="org.eclipse.wildwebdeveloper.debug.IsLaunchableHTMLTester"
+            id="org.eclipse.wildwebdeveloper.propertyTester.isHTMLLaunchable"
+            namespace="org.eclipse.wildwebdeveloper"
+            properties="isHTMLLaunchable"
             type="java.lang.Object">
       </propertyTester>
    </extension>
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/HTMLLaunchableAdapterFactory.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/HTMLLaunchableAdapterFactory.java
@@ -1,0 +1,48 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Andrew Obuchowicz (Red Hat Inc.) 
+ *******************************************************************************/
+package org.eclipse.wildwebdeveloper.debug;
+
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.Adapters;
+import org.eclipse.core.runtime.IAdapterFactory;
+import org.eclipse.debug.ui.actions.ILaunchable;
+import org.eclipse.wildwebdeveloper.debug.firefox.FirefoxRunDebugLaunchShortcut;
+
+// Based off of https://github.com/eclipse/aCute/blob/master/org.eclipse.acute/src/org/eclipse/acute/IsLaunchableHTMLTester.java
+
+public class HTMLLaunchableAdapterFactory implements IAdapterFactory {
+
+	private static ILaunchable DUMMY = new ILaunchable() {
+	};
+
+	@Override
+	public <T> T getAdapter(Object adaptableObject, Class<T> adapterType) {
+		if (!ILaunchable.class.equals(adapterType)) {
+			return null;
+		} 
+			IResource resource = Adapters.adapt(adaptableObject, IResource.class);
+			if  (new FirefoxRunDebugLaunchShortcut().canLaunch(resource.getLocation().toFile())) {
+				return adapterType.cast(DUMMY);
+			
+			
+		}
+
+		return null;
+	}
+
+	@Override
+	public Class<?>[] getAdapterList() {
+		return new Class<?>[] { ILaunchable.class };
+	}
+
+}
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/IsLaunchableHTMLTester.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/IsLaunchableHTMLTester.java
@@ -1,0 +1,33 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Andrew Obuchowicz (Red Hat Inc.) 
+ *******************************************************************************/
+package org.eclipse.wildwebdeveloper.debug;
+
+import org.eclipse.core.expressions.PropertyTester;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.Adapters;
+import org.eclipse.wildwebdeveloper.debug.firefox.FirefoxRunDebugLaunchShortcut;
+
+public class IsLaunchableHTMLTester extends PropertyTester {
+	private static final String PROPERTY_NAME = "isHTMLLaunchable"; //$NON-NLS-1$
+	@Override
+	public boolean test(Object receiver, String property, Object[] args, Object expectedValue) {
+		if (property.equals(PROPERTY_NAME)) {
+			IResource resource = Adapters.adapt(receiver, IResource.class);
+			if (resource == null) {
+				return false;
+			}
+			return new FirefoxRunDebugLaunchShortcut().canLaunch(resource.getLocation().toFile());
+		}
+		return false;
+	}
+}
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxRunDABDebugDelegate.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxRunDABDebugDelegate.java
@@ -89,7 +89,14 @@ public class FirefoxRunDABDebugDelegate extends DSPLaunchDelegate {
 		} catch (IOException | URISyntaxException e) {
 			IStatus errorStatus = new Status(IStatus.ERROR, Activator.PLUGIN_ID, e.getMessage(), e);
 			Activator.getDefault().getLog().log(errorStatus);
-			ErrorDialog.openError(Display.getDefault().getActiveShell(), "Debug error", e.getMessage(), errorStatus); //$NON-NLS-1$
+			Display.getDefault().asyncExec(new Runnable() {
+			@Override
+			public void run() {
+
+					ErrorDialog.openError(Display.getDefault().getActiveShell(), "Debug error", e.getMessage(), errorStatus); //$NON-NLS-1$
+			}
+			});
+
 		}
 	}
 


AObuchow:
Something broke during the rebase. No longer able to launch firefox debugger. Looking into it now.

AObuchow:
> Something broke during the rebase. No longer able to launch firefox debugger. Looking into it now.

Nevermind, was using the wrong target platform. The tests are failing however.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -545,11 +545,25 @@
...
                   </iterate>
                </with></enablement>
          </contextualLaunch>
       </shortcut>
+   </extension>
+    <extension
+         point="org.eclipse.core.runtime.adapters">
+      <factory
+            adaptableType="org.eclipse.core.resources.IResource"
+            class="org.eclipse.wildwebdeveloper.debug.firefox.FirefoxLaunchableAdapterFactory">
```
mickaelistria: This is not really related to firefox, as we'll probably have the same logic for Chrome.
It could be something like "HTMLLaunchableAdapterFactory"

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/Tester.java
```diff
@@ -0,0 +1,47 @@
+package org.eclipse.wildwebdeveloper.debug.firefox;
```
mickaelistria: Missing copyright header.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -545,11 +545,25 @@
                   <iterate>
                      <adapt
                            type="org.eclipse.core.resources.IResource">
+                        <test
+                              forcePluginActivation="true"
+                              property="org.eclipse.wildwebdeveloper.isHTMLProject">
```
mickaelistria: At the moment, I don't think the launch shortcuts allow to properly handle projects, but work fine with files.
So I'd suggest that instead of checking for HTML project, we currently only check for HTML files. We can add projects in a separate commit,

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -560,6 +574,13 @@
             properties="isNodeResource"
             type="java.lang.Object">
       </propertyTester>
+      <propertyTester
+            class="org.eclipse.wildwebdeveloper.debug.firefox.Tester"
+            id="org.eclipse.wildwebdeveloper.propertyTester.isHTMLProject"
```
mickaelistria: Should be more something like `isHTMLLaunchable` so we can reuse it for Chrome support later.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxLaunchableAdapterFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import org.eclipse.core.runtime.IAdapterFactory;
+import org.eclipse.debug.ui.actions.ILaunchable;
+
+public class FirefoxLaunchableAdapterFactory implements IAdapterFactory {
+	
+	private static final ILaunchable DUMMY = new ILaunchable() {
+	};
+
+	@Override
+	public <T> T getAdapter(Object adaptableObject, Class<T> adapterType) {
```
mickaelistria: Usually, first check is for the adpterType

```java
   if (!ILaunchable.class.equals(adapterType)) {
      return null;
   }
```

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxLaunchableAdapterFactory.java
```diff
@@ -0,0 +1,28 @@
...
+		IResource resource = Adapters.adapt(adaptableObject, IResource.class);
+		if (adapterType.equals(ILaunchable.class) && Tester.isHTMLProject(resource.getProject())) {
```
mickaelistria: We also need to return some ILaunchable for HTML files.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/Tester.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.eclipse.core.expressions.PropertyTester;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IAdaptable;
+
+public class Tester extends PropertyTester {
```
mickaelistria: Please give is a more explicit name, like `IsLaunchableHTMLTester`

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/Tester.java
```diff
@@ -0,0 +1,47 @@
...
+		}
+	}
+
+	public static boolean isHTMLProject(IProject p) {
+		if (p == null || !p.isAccessible()) {
+			return false;
+		}
+		try {
+			for (IResource projItem : p.members()) {
+				if (projItem.getName().equals("index.html")) { //$NON-NLS-1$
```
mickaelistria: The launch shortcut doesn't (yet) contain logic about detecting the `index.html`. Is this logic included in the language servers themselves? Ie, if provided a folder, do they automatically start the `index.html` ? That would actually be great and may automatically get things working better in Wild Web Developer.

mickaelistria:
Also, can you please make it 2 commits: one about adding the ILaunchable where is makes sense (so the Debug As context-menu shows in Wild Web Developer), and another one about not showing firefox on non related resources?

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/Tester.java
```diff
@@ -0,0 +1,47 @@
...
+		if (o instanceof IResource) {
+			return (IResource) o;
+		} else if (o instanceof IAdaptable) {
+			return ((IAdaptable) o).getAdapter(IResource.class);
+		} else {
+			return null;
+		}
+	}
+
+	public static boolean isHTMLProject(IProject p) {
```
mickaelistria: Maybe this can simply be `new FirefoxRunDebugLaunchShortcut().canLaunch()` to improve factorization.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/Tester.java
AObuchow: I don't think the debug adapter (I assume this is what you mean by language server) has any logic to automatically start the `index.html`. In my testing, trying to open a folder in Firefox presents the user with a directory explorer.

The logic in `Tester.java` is only used to determine if it makes sense for a file or project to be debuged in Firefox. 

The actual launch shortcut will most likely need logic to find the `index.html` and set that as the `file` parameter in `FirefoxRunDABDebugDelegate`.

org.eclipse.wildwebdeveloper/plugin.xml
AObuchow: I agree. I'll make it check if the user is trying to launch a HTML file rather than checking if the file belongs to a project that contains a `index.html`.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/Tester.java
mickaelistria: > The actual launch shortcut will most likely need logic to find the index.html and set that as the file parameter in FirefoxRunDABDebugDelegate.

Exactly, and I propose we add this logic later to focus on getting the shortcut shown only where it works at the moment (`.html` files).

AObuchow: Agreed. Working on it now :)

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxLaunchableAdapterFactory.java
AObuchow: So I should check `IsLaunchableHTMLTester.isHTMLFile(resource)` and return the ILaunchable?

mickaelistria: Sounds good. Let's try and see ;)

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/Tester.java
AObuchow: Forgot about this change, doing this now.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/IsLaunchableHTMLTester.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+	
+	public static boolean isHTMLFile(IResource resource) {
+		if (resource == null || !resource.isAccessible()) {
+			return false;
+		}
+		return resource.getName().matches((".*\\.html$"));
+	}
+
+	public static boolean isHTMLProject(IProject p) {
```
mickaelistria: As discussed earlier, I think we should get rid of it at the moment since we do not support that case correctly.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxRunDebugLaunchShortcut.java
```diff
@@ -31,6 +32,15 @@ public void configureLaunchConfiguration(File file, ILaunchConfigurationWorkingC
 		wc.setAttribute(FirefoxRunDABDebugDelegate.FILE, file.getAbsolutePath());
 		wc.setAttribute(FirefoxRunDABDebugDelegate.WORKING_DIRECTORY, file.getParentFile().getAbsolutePath());
 	}
+	
+	public static boolean canLaunch(IResource resource) {
```
mickaelistria: This method is already defined in the parent class, no need to reimplement it.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/IsLaunchableHTMLTester.java
```diff
@@ -0,0 +1,33 @@
...
+public class IsLaunchableHTMLTester extends PropertyTester {
+	private static final String PROPERTY_NAME = "isHTMLLaunchable"; //$NON-NLS-1$
+	@Override
+	public boolean test(Object receiver, String property, Object[] args, Object expectedValue) {
+		if (property.equals(PROPERTY_NAME)) {
+			IResource resource = Adapters.adapt(receiver, IResource.class);
+			if (resource == null) {
+				return false;
+			}
+			return FirefoxRunDebugLaunchShortcut.canLaunch(resource);
```
mickaelistria: `new FirefoxRunDebugLaunchShortcut().canLaunch(resource.getLocation().toFile())` would allow to reuse already existing code.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxRunDebugLaunchShortcut.java
AObuchow: Ah my bad, I just realized. Fixing it now.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/HTMLLaunchableAdapterFactory.java
```diff
@@ -0,0 +1,48 @@
...
+	private static ILaunchable DUMMY = new ILaunchable() {
+	};
+
+	@Override
+	public <T> T getAdapter(Object adaptableObject, Class<T> adapterType) {
+		if (!ILaunchable.class.equals(adapterType)) {
+			return null;
+		} else {
+			IResource resource = Adapters.adapt(adaptableObject, IResource.class);
+			if (adapterType.equals(ILaunchable.class) && (new FirefoxRunDebugLaunchShortcut()).canLaunch(resource.getLocation().toFile())) {
```
mickaelistria: You don't need to repeat the condition on adapterType here.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/HTMLLaunchableAdapterFactory.java
```diff
@@ -0,0 +1,48 @@
...
+public class HTMLLaunchableAdapterFactory implements IAdapterFactory {
+
+	private static ILaunchable DUMMY = new ILaunchable() {
+	};
+
+	@Override
+	public <T> T getAdapter(Object adaptableObject, Class<T> adapterType) {
+		if (!ILaunchable.class.equals(adapterType)) {
+			return null;
+		} else {
```
mickaelistria: The `else` here can be removed (`else`s after a `return` are useless).

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/IsLaunchableHTMLTester.java
```diff
@@ -0,0 +1,33 @@
...
+public class IsLaunchableHTMLTester extends PropertyTester {
+	private static final String PROPERTY_NAME = "isHTMLLaunchable"; //$NON-NLS-1$
+	@Override
+	public boolean test(Object receiver, String property, Object[] args, Object expectedValue) {
+		if (property.equals(PROPERTY_NAME)) {
+			IResource resource = Adapters.adapt(receiver, IResource.class);
+			if (resource == null) {
+				return false;
+			}
+			return (new FirefoxRunDebugLaunchShortcut()).canLaunch(resource.getLocation().toFile());
```
mickaelistria: parenthesis here are not necessay.

mickaelistria:
Thanks!

PR #252, opened 2019-09-10
Chrome debugger integration

### LINKED ISSUE #230: Integrate Chrome debugger
I'm going to start working on integrating the chrome debugger as the vscode extension for it has a debuger adapter which is very similar to the node debugger's adapter.

AObuchow:
Fix #230


mickaelistria:
Can we close it as it seems like #256 replaces it?

Issue #238, opened 2019-08-30
Why does wildweb do "/bin/bash -c which node" ?


mbooth101:
I noticed that wildweb does ```/bin/bash -c which node```

Some platforms don't use bash as their default shell, so this could fail out-of-the-box. And AFAIK the "which" command is not even a bash built-in, so invoking the shell seems redundant. It should be sufficient to simply call ```which node```

mickaelistria:
Ok. Can you please submit a PR for that?

PyvesB:
You need to invoke a new shell with `/bin/bash -c which node` as the PATH variable set by the Eclipse will probably not contain the Node executable, therefore `which node` will not return anything. See [this StackOverflow thread](https://stackoverflow.com/a/48831940/3527464) for more information. I would even argue it should be `/bin/bash -c -l which node`, as I've documented in my investigation on [this other thread](https://stackoverflow.com/questions/55436162/programmatically-get-full-paths-of-external-programs-from-within-eclipse).

PyvesB:
Probably not much to add here for the time being, shall we close this issue?

PR #262, opened 2019-09-21
Make bash act as if invoked as a login shell

PyvesB:
This is a follow up to #238. Indeed, Wild Web Developer does not work properly on my macOS Mojave machine. Node.js 10.16.3 is part of my `PATH` variable specified in my _.profile_ file, however, as documented [here](https://stackoverflow.com/a/55716895/3527464), the `-l` option needs to be specified to make sure it is picked up when executing the shell from within Eclipse.

Signed-off-by: PyvesB <PyvesDev@gmail.com>


mickaelistria:
@mbooth101 What do you think about this one?

mickaelistria:
Thanks!

PyvesB:
Tested this morning: Wild Web Developer now works properly on my macOS machine!

PR #269, opened 2019-09-26
Fix default Node.js paths

### LINKED ISSUE #268: Incorrect default Node.js location for macOS
The code that sets the default Node.js location when not found on the macOS PATH does not work as intended:
https://github.com/eclipse/wildwebdeveloper/blob/88cad00cafd837260df54a1020b5434e5700011a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java#L57

`res` cannot be `null` as this point due to the 3 previous lines of code:
```
if (res == null) {
    res = "/path/to/node";
}
```

We should also probably include sensible defaults for other operating systems. I will submit a pull request shortly.

Cheers,

Pyves

PyvesB:
Fixes #268.

Signed-off-by: Pierre-Yves B. <PyvesDev@gmail.com>

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
@@ -12,6 +12,7 @@
  *   Gautier de Saint Martin Lacaze - Issue #55 Warn missing or incompatible node.js
  *   Pierre-Yves B. - Issue #196 NullPointerException when validating Node.js version
  *   Pierre-Yves B. - Issue #238 Why does wildweb do "/bin/bash -c which node" ?
+ *   Pierre-Yves B. - Issue #268 Incorrect default Node.js location for macOS
  *******************************************************************************/
 package org.eclipse.wildwebdeveloper;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
```diff
@@ -85,6 +81,17 @@ public static String which(String program) {
...
 	}
+	
+	private static String getDefaultNodePath() {
+		switch (Platform.getOS()) {
+			case Platform.OS_MACOSX:
+				return "/usr/local/bin/node";
+			case Platform.OS_WIN32:
+				return "C:\\Program Files\\nodejs\\node.exe";
+			default:
+				return "/usr/bin/nodejs";
```
PyvesB: @mickaelistria I believe you use a Linux machine, could you please confirm whether this is a sensible default? My Fedora workstation takes forever to start up... 😄

AObuchow: @PyvesB On my Fedora machine, it's `/usr/bin/node` but I'm not sure if it's always the case

PyvesB: If that's the case, it will probably be similar on Fedora, RHEL and CentOS distributions. I've modified the value accordingly.

Thanks!

mickaelistria:
Squashed and merged as 5af3dcd3f4440418c0f324aa0f72d139773217b0 . Thanks @PyvesB !

PR #284, opened 2019-10-15
Avoid double-testing for node's existance.

mbooth101:
If 'which' yields a positive result then we know that the exe
exists and there no need to test it again for existance.

Signed-off-by: Mat Booth <mat.booth@redhat.com>

### DIFF
--- a/org.eclipse.wildwebdeveloper.feature/feature.xml
+++ b/org.eclipse.wildwebdeveloper.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.7.0.qualifier"
+      version="0.7.1.qualifier"
       provider-name="Eclipse Wild Web Developer project"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
@@ -51,11 +51,12 @@ public class InitializeLaunchConfigurations {
 
 		String res = which("node");
 		if (res == null) {
-			res = getDefaultNodePath();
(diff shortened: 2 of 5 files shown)


mbooth101:
This change is useful for running in sandboxes where node might exist on the sandbox host where, from within the sandbox, you cannot directly observe the exe on the filesystem.

org.eclipse.wildwebdeveloper.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.7.0.qualifier"
+      version="0.7.1.qualifier"
```
mbooth101: I didn't know what the next version should be, so I made the minimum change necessary for the build to complete.

mbooth101:
I think the CI test are unrelated to my change -- I cannot reproduce them locally when I do ```mvn clean verify```

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
```diff
@@ -51,11 +51,12 @@ public static String getNodeJsLocation() {
...
 		if (res == null) {
-			res = getDefaultNodePath();
+			if (Files.exists(Paths.get(getDefaultNodePath()))) {
+				res = getDefaultNodePath();
+			}
 		}
 
-		if (Files.exists(Paths.get(res))) {
-
+		if (res != null) {
```
mickaelistria: Can getDefaultNodePath return invalid location in some cases (ie node has moved or something like that? If so, then it could make sense to keep the check.

mbooth101: Yes, but I didn't get rid of the check -- I just moved it up a bit so it's only run if ```which("node")``` returns null.

mickaelistria: Ok

mickaelistria:
Thanks Matt!

PR #299, opened 2019-11-20
Don't use deprecated getWorkbench method.

akurtakov:

### DIFF
--- a/org.eclipse.wildwebdeveloper.feature/feature.xml
+++ b/org.eclipse.wildwebdeveloper.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.7.2.qualifier"
+      version="0.7.3.qualifier"
       provider-name="Eclipse Wild Web Developer project"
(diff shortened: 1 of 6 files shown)


org.eclipse.wildwebdeveloper.feature/feature.xml
```diff
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.7.2.qualifier"
+      version="0.7.3.qualifier"
```
PyvesB: Given that version 0.8.0 was just officially released, why are there older version numbers here?

akurtakov: @mickaelistria would have to say that :)

akurtakov: I'm going to merge this PR as it's semver related. E.g. version 0.8 is due to the API/extension point added in the xml bundle. I still think that feature version should reflect that though.

mickaelistria:
Right, the feature version should be 0.8.0 for the release, and 0.8.1 now. I just forgot it.

PR #304, opened 2019-11-27
Add ESLint Support

AObuchow:


org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -127,6 +127,7 @@
 								<arg>debugger-for-chrome-4.11.7.tgz</arg>
 								<arg>@angular/language-service@9.0.0-next.8</arg>
 								<arg>@angular/language-server@0.900.0-next.1</arg>
+								<arg>typescript-eslint-language-service@1.4.0</arg>
```
mickaelistria: I opened CQ http://dev.eclipse.org/ipzilla/show_bug.cgi?id=21204 for this

org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestESLint.java
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	@After
+	public void deleteProjectAndCloseEditors() throws Exception {
+		PlatformUI.getWorkbench().getActiveWorkbenchWindow().getActivePage().closeAllEditors(false);
+		this.project.delete(true, null);
+	}
+
+	@Test
+	public void testESLintDiagnostics() throws Exception {
```
mickaelistria: I tried to run this scenario manually: Import project, run `npm install`, open .ts file, copy it to another file... I could see some plain TypeScript diagnotics because of duplicated variables, but nothing from ESLint.

AObuchow: What do you mean by "copy it to another file"? A caveat of ESLint is that the files to be linted must reside within the same directory (or in a child directory) that contains the `.eslintrc`, `tsconfig.json` and the required `node_modules`. If you open a .ts file that isin't in that directory hierarchy, it won't be linted.

Also, did you see if the test was failing?

AObuchow: I tried downgrading my `typescript` version to the one that's packaged with Wild Web Developer and it seems to not work with ESLint. Will have to bump up the typescript version.

mickaelistria: I created a  new file `ESLint2.ts` just beside `ESLint.ts` and coped content and didn't get any linting diagnostic.
Running the test (from IDE as JUnit Plugin Test) fails for me. At the moment, it opens the external text editor, but I'm not sure it's related.

mickaelistria: So I could get the test opening the right editor, but it's now failing "properly" with _Diagnostic not published_ running both from IDE and with `mvn clean verify`, similarly to how it happens on CI.
Maybe the difference is cause by different node versons. What's yours? mine is `v10.16.3`.

mickaelistria:
> I tried downgrading my typescript version to the one that's packaged with Wild Web Developer and it seems to not work with ESLint. Will have to bump up the typescript version.

Ok, can you please rebase on top of https://github.com/eclipse/wildwebdeveloper/pull/306 to try that?

AObuchow:
I'm also working on adding usage notes in the `README.md`.

AObuchow:
The tests are failing on CIBot, will have to figure out why.

PR #305, opened 2019-11-27
Launch tsserver from same node process

### LINKED ISSUE #302: tsserver fails to start if node isn't on PATH
This is possibly a dup of #174.

I don't have a global installation of NodeJS, so I've set org.eclipse.wildwebdeveloper.nodeJSLocation in eclipse.ini instead. tsserver fails to start with this error message in the JS-TS console:

> [Error]	[tsserver] 'node' is not recognized as an internal or external command,
operable program or batch file.

After some investigation, I discovered the reason is that by default, typescript-language-server tries to launch tsserver using child_process#spawn(). However, because that might mean that tsserver runs using a different NodeJS than t-l-s itself, the t-l-s devs also added a second option. If tsserverPath is provided and is a path to a .js file, then t-l-s starts tsserver using child_process#fork() instead, which means tsserver runs using the same NodeJS as t-l-s and node does not need to be on PATH. See https://github.com/theia-ide/typescript-language-server/pull/88.

So, Wild Web Developer should provide tsserverPath when starting t-l-s so that tsserver will use the same NodeJS as everything else.

AObuchow:
Fix #302


mickaelistria:
Thanks @AObuchow !

PR #306, opened 2019-11-28
Update language servers

mickaelistria:
Signed-off-by: Mickael Istria <mistria@redhat.com>

PR #307, opened 2019-11-28
Add eslint support 

AObuchow:


mickaelistria:
ESLint tests fail for me similarly to how they fail on CI.

AObuchow:
I enabled logging of the tsserver to try and see if there's any hints as to why the tests are failing.

AObuchow:
It turns out the dependencies (`eslint@6.0.0, @typescript-eslint/eslint-plugin@2.7.0, @typescript-eslint/parser@2.7.0`) need to be installed in the `language_servers` in order for the typescript-eslint-language-service plugin to be loaded correctly (despite the fact the test project will have them).

The TestDebug.testFindThreadsAndHitsBreakpoint is still failing, although I'm not sure if this is related to my patch.

AObuchow:
> The TestJSON is also failing for me locally, and the JSON language server seems actually broken by that change: I cannot use it manually. The reason is that the serverCapabilities aren't returned properly by the LS, instead, the LS says `{"jsonrpc":"2.0","id":"1","error":{"code":-32603,"message":"Request initialize failed with message: Cannot read property 'provideFormatter' of undefined"}}`

Ok, I'm going to look into it right now.

AObuchow:
It looks like the newest version of the JSON LS (`1.2.2`) expects intialization options. The JSON LS was updated to `1.2.2` in https://github.com/eclipse/wildwebdeveloper/pull/306. See https://github.com/eclipse-theia/theia/issues/6393  & https://github.com/vscode-langservers/vscode-json-languageserver-bin/issues/3.

What confuses me is why the JSON LS tests passed during the build of https://github.com/eclipse/wildwebdeveloper/pull/306 but not for this change (or why they passed at all).

AObuchow:
I reverted https://github.com/eclipse/wildwebdeveloper/pull/306 but left some of the language server updates that should be okay (eg. Angular) - this might cause problems however so it might be best to leave all language servers which aren't relevant to this patch alone.

Also the XML test for `testLanguageServer` was giving a ```java.lang.ClassCastException: org.eclipse.wst.xml.ui.internal.tabletree.XMLMultiPageEditorPart cannot be cast to org.eclipse.ui.texteditor.ITextEditor``` for some reason. We'll see if it occurs on the CI.

mickaelistria:
> I reverted #306 but left some of the language server updates that should be okay (eg. Angular) - this might cause problems however so it might be best to leave all language servers which aren't relevant to this patch alone.

If it cause problem, just ignore all other language servers, and upgrade only the dependencies that are necessary for ESLint to work.
This will also be easier in term of IP, as the CQ for language servers is taking some time, while ESLint is already approved. The least dependencies we change, the faster it will be to get all CQs approved and the patch merged.

> Also the XML test for testLanguageServer was giving a java.lang.ClassCastException: org.eclipse.wst.xml.ui.internal.tabletree.XMLMultiPageEditorPart cannot be cast to org.eclipse.ui.texteditor.ITextEditor for some reason. We'll see if it occurs on the CI.

This is because in your IDE/Launch Configuration, webtools XML editor is included and it takes priority of Generic Editor. On CI and in the Maven build, WebTools isn't part of the target-platform, so this shouldn't happen.

AObuchow:
> This will also be easier in term of IP, as the CQ for language servers is taking some time, while ESLint is already approved. The least dependencies we change, the faster it will be to get all CQs approved and the patch merged.

This is enough of a reason for me to leave the other language servers unchanged.

> 
> > Also the XML test for testLanguageServer was giving a java.lang.ClassCastException: org.eclipse.wst.xml.ui.internal.tabletree.XMLMultiPageEditorPart cannot be cast to org.eclipse.ui.texteditor.ITextEditor for some reason. We'll see if it occurs on the CI.
> 
> This is because in your IDE/Launch Configuration, webtools XML editor is included and it takes priority of Generic Editor. On CI and in the Maven build, WebTools isn't part of the target-platform, so this shouldn't happen.

Ok good to know - this is due to having m2e in the TP (related to XML LS work)

AObuchow:
Also do we need a version bump? Maybe +0.1.0?

mickaelistria:
No, it should be fine for versions. The build is configured so that if we need a version bump, it should fail with an explicit error message.

AObuchow:
All tests are now passing. I had to do a +0.1.0 for the build on the testing plugin for the build to be successful.

mickaelistria:
@AObuchow I've added you as a cc. of CQ https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21224 so you'll receive notifications. If this gets marked as fixed, then feel free to merge the patch whenever you think it's ready (I see a test failing, I'll retrigger to see it green).

AObuchow:
> @AObuchow I've added you as a cc. of CQ https://dev.eclipse.org/ipzilla/show_bug.cgi?id=21224 so you'll receive notifications. If this gets marked as fixed, then feel free to merge the patch whenever you think it's ready 

Ok sounds good. 

>(I see a test failing, I'll retrigger to see it green)

I see that as well, I'm not sure why more tests are now failing when they were passing 2 builds ago (no code change IIRC). I'll have to investigate.

AObuchow:
@mickaelistria The latest build passed all tests. I guess the previous 2 builds failed due to timing issues.

mickaelistria:
Despite the test failures on CI, I could test the testTSFile scenario locally from this commit (after a build and installation from p2 repo in a vanilla IDE) and see it work. Other failures seem to happen only sporadically.
I think that this patch is good, but legacy tests should be improved at some point. I'm merging it.

PR #310, opened 2019-12-02
Call DebugUITools.launch from the UI thread

### LINKED ISSUE #309: Launch called from the wrong thread?
Hi there!

I noticed that the following line of code fires the launch asynchronously:
https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractDebugAdapterLaunchShortcut.java#L157

However, the Javadocs for `DebugUITools.launch` state the following: 

> This method must be called in the UI thread.

I'm unsure what the consequences of calling this method from a different thread are, but this is probably something worth looking into.

PyvesB:
Closes  #309. Note that underneath the hood, the program is launched in a background job, so this change should not impact UI responsiveness.

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractDebugAdapterLaunchShortcut.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractDebugAdapterLaunchShortcut.java
@@ -6,12 +6,14 @@
  * which is available at https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
+ * 
+ * Contributors:
+ *   Pierre-Yves B. - Issue #309 Launch called from the wrong thread
  *******************************************************************************/
 package org.eclipse.wildwebdeveloper.debug;
 
 import java.io.File;
 import java.util.Arrays;
-import java.util.concurrent.CompletableFuture;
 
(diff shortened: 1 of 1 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractDebugAdapterLaunchShortcut.java
```diff
@@ -154,7 +156,7 @@ public void launch(IEditorPart editor, String mode) {
 
 	private void launch(String mode, ILaunchConfiguration[] configurations) {
 		if (configurations.length == 1) {
-			CompletableFuture.runAsync(() -> DebugUITools.launch(configurations[0], mode));
+			DebugUITools.launch(configurations[0], mode);
```
mickaelistria: IIRC, it was important that this method would be non-blocking and that's why it got wrapped in a Future.
I think calling `Display.getDefault().asyncExec(() -> DebugUITools.launch(configurations[0], mode))` is safer here.

PyvesB: Done.

mickaelistria:
Test failures are not related. So it's good to merge.

PR #313, opened 2019-12-05
Add m2e dependencies onto XML LS classpath

AObuchow:
Work in progress

### DIFF
--- a/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClasspathExtensionProvider.java
+++ b/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClasspathExtensionProvider.java
@@ -1,0 +1,25 @@
+/*******************************************************************************
+ * Copyright (c) 2019 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Andrew Obuchowicz (Red Hat Inc.) - initial implementation
+ *******************************************************************************/
+package org.eclipse.wildwebdeveloper.xml;
+
+import java.io.File;
+import java.util.List;
+import java.util.function.Supplier;
+
+/**
+ * @since 0.6.0
+ *
+ */
+public interface XMLLSClasspathExtensionProvider extends Supplier<List<File>> {
+
+}
--- a/org.eclipse.wildwebdeveloper.xml/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper.xml/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: XML support using Language Server
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper.xml;singleton:=true
-Bundle-Version: 0.5.1.qualifier
+Bundle-Version: 0.6.0.qualifier
 Bundle-Vendor: Eclipse Wild Web Developer project
 Automatic-Module-Name: org.eclipse.wildwebdeveloper.xml
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
--- a/org.eclipse.wildwebdeveloper.xml/pom.xml
+++ b/org.eclipse.wildwebdeveloper.xml/pom.xml
@@ -7,7 +7,7 @@
 		<version>0.5.0-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-plugin</packaging>
-	<version>0.5.1-SNAPSHOT</version>
+	<version>0.6.0-SNAPSHOT</version>
 
 	<build>
 		<plugins>
--- a/org.eclipse.wildwebdeveloper.xml/schema/xmllsExtension.exsd
+++ b/org.eclipse.wildwebdeveloper.xml/schema/xmllsExtension.exsd
@@ -7,6 +7,8 @@
       </appinfo>
       <documentation>
          This extension point is used to contribute LSP4XML extension jars to extend the functionality of the XML language server.
+
+Jars can be contributed as an extension resource (by pointing to the path of the jar) or programmatically, by providing an implementation of the &lt;b&gt;XMLLSClasspathExtensionProvider&lt;/b&gt; interface.
       </documentation>
    </annotation>
 
@@ -18,7 +20,8 @@
       </annotation>
       <complexType>
          <sequence>
-            <element ref="jar" minOccurs="1" maxOccurs="unbounded"/>
+            <element ref="jar" minOccurs="0" maxOccurs="unbounded"/>
+            <element ref="classpathExtensionProvider" minOccurs="0" maxOccurs="unbounded"/>
          </sequence>
          <attribute name="point" type="string" use="required">
             <annotation>
@@ -49,13 +52,28 @@
 
    <element name="jar">
       <complexType>
-         <attribute name="path" type="string" use="required">
+         <attribute name="path" type="string">
             <annotation>
                <documentation>
                   
                </documentation>
                <appinfo>
                   <meta.attribute kind="resource"/>
+               </appinfo>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="classpathExtensionProvider">
+      <complexType>
+         <attribute name="provider" type="string">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+               <appinfo>
+                  <meta.attribute kind="java" basedOn=":org.eclipse.wildwebdeveloper.xml.XMLLSClasspathExtensionProvider"/>
                </appinfo>
             </annotation>
          </attribute>
@@ -76,11 +94,22 @@
          <meta.section type="examples"/>
       </appinfo>
       <documentation>
-         &lt;br&gt; &amp;lt;extension &lt;br&gt;
-        &lt;span style=&quot;margin-left:2em&quot;&gt; point=&amp;quot;org.eclipse.wildwebdeveloper.xml.XMLExtensions&amp;quot;&amp;gt;&lt;br&gt;
+         &lt;br&gt;&lt;b&gt;&lt;u&gt;Providing a jar as a resource&lt;/u&gt;&lt;/b&gt;&lt;br&gt;
+
+&lt;br&gt; &amp;lt;extension &lt;br&gt;
+        &lt;span style=&quot;margin-left:2em&quot;&gt;point=&amp;quot;org.eclipse.wildwebdeveloper.xml.xmllsExtension&amp;quot;&amp;gt;&lt;br&gt;
      &lt;span style=&quot;margin-left:1em&quot;&gt;&amp;lt;jar&lt;br&gt;
            &lt;span style=&quot;margin-left:3em&quot;&gt; path=&amp;quot;/path/to/extension.jar&amp;quot;&amp;gt;&lt;br&gt;
     &lt;span style=&quot;margin-left:1em&quot;&gt;&amp;lt;/jar&amp;gt;&lt;br&gt;
+    &amp;lt;/extension&amp;gt;
+&lt;br&gt;
+&lt;br&gt;    
+&lt;b&gt;&lt;u&gt;Providing a jar programmatically&lt;/u&gt;&lt;/b&gt;
+&lt;br&gt; &amp;lt;extension &lt;br&gt;
+     &lt;span style=&quot;margin-left:2em&quot;&gt; point=&amp;quot;org.eclipse.wildwebdeveloper.xml.xmllsExtension&amp;quot;&amp;gt;&lt;br&gt;
+           &lt;span style=&quot;margin-left:1em&quot;&gt;&amp;lt;classpathExtensionProvider&lt;br&gt;
+                 &lt;span style=&quot;margin-left:3em&quot;&gt;provider=&amp;quot;org.eclipse.wildwebdeveloper.xml.XMLLSClasspathExtensionProvider&amp;quot;&amp;gt;&lt;br&gt;
+           &lt;span style=&quot;margin-left:1em&quot;&gt;&amp;lt;/classpathExtensionProvider&amp;gt;&lt;br&gt;
     &amp;lt;/extension&amp;gt;
       </documentation>
    </annotation>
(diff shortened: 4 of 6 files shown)


angelozerr:
@AObuchow could you explain more the goal of this PR please? 

I think you want to consume M2E API in the LSP4XML extensions. I though about this idea but I would to consume the LSP4XML Maven extension inside vscode too with vscode-java (and JDT LS which provides too M2 support).

If you consume directly M2 API, you will load twice the Maven project model (one by M2E and one by LSP4XML Maven extension). I think Maven project model should be loaded once.

So my idea was to have a LSP4XML  Maven extension which delegates Maven feature to the client (I think @mickaelistria will not like this idea :)). Today it's not possible but there is @martinlippert who could be interested by this feature https://github.com/angelozerr/lsp4xml/issues/596

mickaelistria:
You got it right Angelo, I don't support anything that involves writing more client specific and non-portable code ;)
About the change, yes, the Maven POM models would be loaded twice; but they actually already loaded twice in m2e IIRC: the editor does work with a copy of the pom model that can be different from the one used by the internal state for classpath resolution; so this wouldn't make much difference.
In the end, we're also duplicating a full JVM execution, full XML parsers and so on... I think the idea (and value) of separation/isolation comes with the drawback that we duplicate work in different process, but still, it's IMO what should be preferred rather than building too complex integrations. Moreover some people may like to have the XML-LS+Maven support running without having the Java support running and vice-version.

org.eclipse.wildwebdeveloper.xml/plugin.xml
```diff
@@ -144,5 +145,11 @@
             icon="icons/xmlEditorIcon.png">
       </icon>
    </extension>
+   <extension
+         point="org.eclipse.wildwebdeveloper.xml.xmllsClasspathExtensionProvider">
+      <ClassPathExtensionProvider
```
mickaelistria: elements should use lowercase (`classpathExtensionProvider`)

org.eclipse.wildwebdeveloper.xml/plugin.xml
```diff
@@ -144,5 +145,11 @@
...
+            Provider="org.eclipse.wildwebdeveloper.xml.MavenArtifactClassPathProvider">
```
mickaelistria: lower case: `provider`

org.eclipse.wildwebdeveloper.xml/schema/xmllsClasspathExtensionProvider.exsd
```diff
@@ -0,0 +1,102 @@
...
+         </attribute>
+      </complexType>
+   </element>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="since"/>
+      </appinfo>
+      <documentation>
+         [Enter the first release in which this extension point appears.]
```
mickaelistria: 0.9.0 IIRC.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/BundleResolver.java
```diff
@@ -0,0 +1,61 @@
...
+	private BundleResolver() {
+		// Utility class, not meant to be instantiated
+	}
+
+	public static Bundle getBundle(String bundleID) {
+		Bundle bundle = Platform.getBundle(bundleID);
+		// TODO: Decide if we really need to have the bundle activated
+		if (bundle.getState() != Bundle.ACTIVE) {
+			try {
+				bundle.start();
```
mickaelistria: Do you need to start the bundle to get its resources? I'm don't think so.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/MavenArtifactClassPathProvider.java
```diff
@@ -0,0 +1,12 @@
+package org.eclipse.wildwebdeveloper.xml;
+
+import java.io.File;
+
+public class MavenArtifactClassPathProvider implements XMLLSClasspathExtensionProvider {
```
mickaelistria: This should be in the m2e patch that's currently on Gerrit.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/MavenArtifactClassPathProvider.java
```diff
@@ -0,0 +1,12 @@
+package org.eclipse.wildwebdeveloper.xml;
```
mickaelistria: Missing legal header.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClassPathExtensionRegistry.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+
+public class XMLLSClassPathExtensionRegistry {
```
mickaelistria: Actually, I'm not sure you really need to keep a copy of the model here. Maybe it's easier and more powerful and cheap enough to query the extension registry everytime it's needed. WDYT?

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClassPathExtensionRegistry.java
```diff
@@ -0,0 +1,61 @@
+package org.eclipse.wildwebdeveloper.xml;
```
mickaelistria: legal header

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClasspathExtensionProvider.java
```diff
@@ -0,0 +1,8 @@
+package org.eclipse.wildwebdeveloper.xml;
```
mickaelistria: Legal header.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClasspathExtensionProvider.java
```diff
@@ -0,0 +1,8 @@
...
+
+import java.io.File;
+import java.util.function.Supplier;
+
+public interface XMLLSClasspathExtensionProvider extends Supplier<File> {
```
mickaelistria: add a `@since 0.9.0` annotation here.

AObuchow:
> I think you want to consume M2E API in the LSP4XML extensions. I though about this idea but I would to consume the LSP4XML Maven extension inside vscode too with vscode-java (and JDT LS which provides too M2 support).

This is correct, the idea is to consume the Maven API in LSP4XML maven extension. If I'm not mistaken, the Maven API jars should also be accesible from vscode-java, which could add the jars onto the XML LS classpath in vscode.

> So my idea was to have a LSP4XML Maven extension which delegates Maven feature to the client (I think @mickaelistria will not like this idea :)). Today it's not possible but there is @martinlippert who could be interested by this feature [angelozerr/lsp4xml#596](https://github.com/angelozerr/lsp4xml/issues/596)

This is an interesting proposition and I'll be following it closely (communication between 2 LS'), but I agree with Mickael that client-specific code doesn't fit with the architecture of Wild Web Developer. Perhaps later on (eg. inter-LS-communication becomes part of the LS spec and we can assume all clients can support it) this should be considered.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClassPathExtensionRegistry.java
AObuchow: I'm not entirely sure I understand. Do you mean we should do without the "sync extension registry" pattern here? That should be ok, as in this implementation we only need to query the extension registry when the XML LS is launched.

I looked into adding onto the classpath dynamically (after the LS would be launched) and it seems to be possible but [involves using class loaders and reflection.](https://community.oracle.com/thread/1239849) Im not sure if there are any use cases that would merit dynamically adding onto the classpath. Thus it should be safe to say that the xtension registry is only needed when launching the XML LS.

mickaelistria: > Do you mean we should do without the "sync extension registry" pattern here?

Yes, but actually, we can just keep it as it at the moment; unless you think it's worth changing it right now.

> I looked into adding onto the classpath dynamically (after the LS would be launched) 

Adding to classpath after load is too hard, and as you identified, there isn't use-case for it.
I don't think we need more dynamic handling. My suggestion was more about whether we do really need an sync'd cache for this case or whether just querying the extension registry on every start of the LS would be simpler.

AObuchow: Ok sounds good. Lets keep it for the moment (and fix it before the merge) but I agree there's no reason to cache the extension registry. This is probably also true for the xmllsExtension/XMLExtension registry (but that's another topic).

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/MavenArtifactClassPathProvider.java
AObuchow: I kept it in the most recent commit but am going to remove it now and add it to the gerrit m2e patch.

martinlippert:
> About the change, yes, the Maven POM models would be loaded twice; but they actually already loaded twice in m2e IIRC: the editor does work with a copy of the pom model that can be different from the one used by the internal state for classpath resolution; so this wouldn't make much difference.
> In the end, we're also duplicating a full JVM execution, full XML parsers and so on... I think the idea (and value) of separation/isolation comes with the drawback that we duplicate work in different process, but still, it's IMO what should be preferred rather than building too complex integrations. Moreover some people may like to have the XML-LS+Maven support running without having the Java support running and vice-version.

While I agree in general, I also think that specific tasks should probably be done only once. The loading and resolution of the Maven model is a good example to discuss in this setting, I think.

For the Spring Boot language server, we did our own Maven model building in the beginning and we moved away from that for various reasons:

- it turns out to be quite difficult to ensure that the project identification and classpath resolution is always the exact same as the one that the Java tooling produces. It resulted in the Java tooling and the Spring tooling doing their work on different assumptions about the project structure and therefore producing confusing results.

- resolving the Maven model and doing classpath resolution turned out to be a quite expensive and resource intensive task, so doing it twice or even more (e.g. Java + XML + Spring) is extremely resource intensive (CPU + memory)

- since this work is quite resource intensive, you would need to implement caches/optimizations along the way (trust me, you will definitely need this) - so every language server would need to implement its own mechanism for this - and it is questionable whether every language server should re-implement this

- shipping the Maven libs as part of the language server increases the size of the language server package, classloading necessary, native/image size, etc.

I think this is not always a black or white decision. We found specific tasks to be duplicated in the language servers makes a lot of sense. But the project/classpath resolution for Maven is, from my point of view, a good example for NOT doing it in every language server again... :-)

angelozerr:
Thanks so much @martinlippert for your great feedback! 

If I understand your comments you mean that the LSP4XML maven extension should delegate the call of M2 maven features to the LSP language client (in Eclipse IDE context, language client will be an implementation of LSP4E LanguageClientImlpl, and in vscode TypeScript language client will delegate the M2 features to JDT LS), is that?

mickaelistria:
> But the project/classpath resolution for Maven is, from my point of view, a good example for NOT doing it in every language server again... :-)

I agree. However, the focus here is only pom.xml edition (ie read settings, parse pom file, and interact with remote and local repository). The project/classpath resolution seems to be a different enough line of work at the moment.
Anyway, we need such classpath extensibility for XML-LS (even for other use-cases), so the proposed extension still is interesting. We'll probably just try it until we reach the limit of what can be done properly without sharing more than the jars.

mickaelistria:
@AObuchow Can you please extract in a separate commit the new extension element?
We now have https://git.eclipse.org/r/#/c/152751/ with the Maven part, so it seems better removing it from this PR so we can merge the new extension element.

AObuchow:
> @AObuchow Can you please extract in a separate commit the new extension element?
> We now have https://git.eclipse.org/r/#/c/152751/ with the Maven part, so it seems better removing it from this PR so we can merge the new extension element.

Sounds good, will do.

I also think it might be better for the extension point to support `Supplier<List<File>>` instead of `Supplier<File>` so that multiple jars can be provided onto the XML LS classpath from a single contributing class. This is done in https://github.com/AObuchow/wildwebdeveloper/commit/3a2b6e5ba21457264d0cbd1d1d049a286c3ee173 (but it needs to be reworked a bit).

If you think this is okay, I'll update this PR with the commit linked above.

mickaelistria:
OK for me.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/XMLLSClassPathExtensionRegistry.java
```diff
@@ -0,0 +1,76 @@
...
+		this.extensions.entrySet().stream()
+				.map(Entry<IConfigurationElement, XMLLSClasspathExtensionProvider>::getValue)
+				.map(XMLLSClasspathExtensionProvider::get)
+				.forEach(list -> list.forEach(jar -> classpathExtensions.add(jar.getAbsolutePath())));
+
+		return classpathExtensions;
+	}
+
+	private void sync() {
+		Set<IConfigurationElement> toRemoveExtensions = new HashSet<>(this.extensions.keySet());
```
AObuchow: I'll move this logic (line 52 to 69) to the `getXMLLSClassPathExtensions` function? (As we agreed there's no need for the sync pattern here)

mickaelistria: Both works. I let you decide ;)

org.eclipse.wildwebdeveloper.xml/plugin.xml
```diff
@@ -2,6 +2,7 @@
 <?eclipse version="3.4"?>
 <plugin>
    <extension-point id="xmllsExtension" name="%ExtPoint.xmllsExtension" schema="schema/xmllsExtension.exsd"/>
+   <extension-point id="xmllsClasspathExtensionProvider" name="%ExtPoint.xmllsClasspathExtensionProvider" schema="schema/xmllsClasspathExtensionProvider.exsd"/>
```
mickaelistria: I think it would rather be the same extension point, but allowing different children element (`jar` and `classpathProvider`). As both serve the same goal, it's more consistent to use the same extension point.

AObuchow: Agreed, I'll try to fix this.

mickaelistria:
@AObuchow it seems to me this is ready to be merged. Is it OK on your side?

AObuchow:
> @AObuchow it seems to me this is ready to be merged. Is it OK on your side?

Yes should be good to merge :)

mickaelistria:
Thanks a lot @AObuchow !

PR #316, opened 2019-12-28
Add extended JSON schema support via JSON preference page

ndoschek:
Thank you for your hint to implement the JSON schema associations via preference pages (see PR #166 ).
I added a new preference page to associate JSON schemata to file patterns. The user can now define the schema location either by adding a valid URL or a local file path.
Also, the existing schema associations are added as default entries (except for the schema 'typing.json', as mentioned in my initial PR comment).

I am looking forward to your opinion on this change.
Best regards!


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSONPreferenceInitializer.java
```diff
@@ -0,0 +1,40 @@
...
+	public static final String JSON_SCHEMA_PREFERENCE = "wildwebdeveloper.json.schema";
+
+	@Override
+	public void initializeDefaultPreferences() {
+		STORE.setDefault(JSON_SCHEMA_PREFERENCE, getDefaultJSONSchema());
+	}
+
+	private static String getDefaultJSONSchema() {
+		JsonObject schemaJson = new JsonObject();
+		schemaJson.addProperty("/.bower.json", "http://json.schemastore.org/bower");
```
mickaelistria: Shouldn't we declare those file names and patterns as content types (children of json contenttype) to the whole IDE, and then have the preference associate a content-type with a schema URL ?
I believe reusing the content type mechanisms allows more powerful features usually.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSONPreferenceInitializer.java
```diff
@@ -0,0 +1,40 @@
...
+		schemaJson.addProperty("/.bowerrc", "http://json.schemastore.org/bowerrc");
```
mickaelistria: I think the `.bowerrc` wouldn't be recognized as a json file anyway. We first need to define it as a child of the json content type.

angelozerr:
IMHO I think json support should be hosted in a new `org.eclipse.wildwebdeveloper.json` like we have for XML. I think this PR it's a good canditate to create this project.

mickaelistria:
> I think this PR it's a good canditate to create this project.

I disagree. Let's get the feature in, and discuss how to factor stuff separately. The contributor shouldn't have to care about this as a prerequisite to get the contribution merged.
@ndoschek Please stick to current code and internal review, no need for you to bother with creating a separate bundle.

ndoschek:
Thank you for your feedback!

I agree and added content types for the default associations as children of the JSON content type.
Now a content type is associated with a schema location in the preferences and each contained pattern is passed to the LS as a separate schema association entry.
If it is acceptable for you, I would still allow adding user specific JSON schema associations for custom file patterns (as in other editors like VSCode).

Unfortunately, I came across a problem defining content types via file-patterns additionally to file-names or file-extensions.
For example, 'tslint.json' is recognized correctly as TSLint Configuration, but 'error.tslint.json' is matched to a simple JSON file.
I found a bug description (https://bugs.eclipse.org/bugs/show_bug.cgi?id=535510), which addresses this problem.
For the schema support via the LS it is not blocking, as each pattern is passed to the LS and is therefore supported. But it might still be confusing for the attentive user. Do you have an idea how to overcome this issue in the meantime?

Additionally, I added new content types (i.e. typings.json, tslint.json, eslint.json, lerna.json) and two editor icons (i.e. bowerEditorIcon.png and npmEditorIcon.png) and fixed a small bug in the YAML preference page schema table.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -53,6 +128,46 @@
...
+            id="org.eclipse.wildwebdeveloper.json">
+      </contentTypeMapping>
+      <contentTypeMapping
+            contentType="org.eclipse.wildwebdeveloper.json.tslint"
+            id="org.eclipse.wildwebdeveloper.json">
+      </contentTypeMapping>
+      <contentTypeMapping
+            contentType="org.eclipse.wildwebdeveloper.json.eslintrc"
+            id="org.eclipse.wildwebdeveloper.json">
+      </contentTypeMapping>
```
mickaelistria: As all those content-types are children of the already associated org.eclipse.wildwebdeveloper.json content-type, you shouldn't have to add new associations, the existing one from the base type should be enough.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -65,6 +180,46 @@
...
+            scopeName="source.json">
+      </scopeNameContentTypeBinding>
+      <scopeNameContentTypeBinding
+            contentTypeId="org.eclipse.wildwebdeveloper.json.tslint"
+            scopeName="source.json">
+      </scopeNameContentTypeBinding>
+      <scopeNameContentTypeBinding
+            contentTypeId="org.eclipse.wildwebdeveloper.json.eslintrc"
+            scopeName="source.json">
+      </scopeNameContentTypeBinding>
```
mickaelistria: Same as above, association from the base type org.eclipse.wildwebdeveloper.json should be enough and not require extra definition here.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -73,6 +228,46 @@
...
+            path="language-configurations/json/language-configuration.json">
+      </languageConfiguration>
+      <languageConfiguration
+            contentTypeId="org.eclipse.wildwebdeveloper.json.tslint"
+            path="language-configurations/json/language-configuration.json">
+      </languageConfiguration>
+      <languageConfiguration
+            contentTypeId="org.eclipse.wildwebdeveloper.json.eslintrc"
+            path="language-configurations/json/language-configuration.json">
+      </languageConfiguration>
```
mickaelistria: And same thing ;)

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSonLanguageServer.java
```diff
@@ -78,11 +82,50 @@ public void handleMessage(Message message, LanguageServer languageServer, URI ro
...
-		Map<String, String> conversion = new Gson().fromJson(schemaString, new TypeToken<HashMap<String, String>>(){}.getType());
+		Map<String, String> conversion = new Gson().fromJson(schemaString, new TypeToken<HashMap<String, String>>() {
+		}.getType());
 
-		Map<String, List<String>> associations = new HashMap<>();
-		conversion.forEach((key, value) -> associations.put(key, Arrays.asList(value)));
+		Map<String, List<String>> schemaAssociations = new HashMap<>();
 
-		return associations;
+		IExtensionPoint point = Platform.getExtensionRegistry()
```
mickaelistria: Instead of looking directly in the plugin.xml, it would work better to use the content-type manager (`Platform.getContentTypeManager()`) to find out which content-types have json as basetype and read their `getFileSpec()` to infer the names and patterns.
The content-type registry can then be customized by other plugins or by end-users to add patterns and so so, which would be automatically passed to the Language Server.

ndoschek:
Thank you for the further hints! I used the content type manager to resolve file associations and linked the content-type preference page on the schema associations page.
I have another question: Is there a simpler way to retrieve (or set?) the ID of a custom content-type in the workspace? I created a custom content-type, and found its ID to be 'userCreated*'. Thank you in advance!

mickaelistria:
Thanks @ndoschek and sorry for the delay in response (I've been a bit busy this last months), I'll review that later today hopefully.
About your question, you can retrieve the id of user-defined content-type with the regular API, with things like `Platform.getContentTypeManager().findContentTypeFor("blah").getId()`. It's indeed usually `userDefined<someRandomishNumber>`. Those ids are not meant to be written "manually", they have to retrieved through the API.

AObuchow:
Could #430 be of use for this topic? I might have an interest in a schema preference page

ndoschek:
Hello again!
I updated my change to the latest master and introduced a preference page for associating content types to JSON schemas.
Therefore I removed the existing extension point for JSON schemas. If you would rather keep it I can update my change and can try to include those schema associations as well.
Looking forward to hear your opinion on this!

mickaelistria:
That's really good progress! I'm wondering whether it wouldn't make sense to merge the JSON and the YAML schemas page into a single one and even use the same preference. What do you think about it?

ndoschek:
Thank you for your response, that sounds good to me as well!
So therefore I would propose to have a single preference page (named `Schema associations`) as a subpage of the `General` preferences).
This page will allow to associate contentTypes only with base types JSON or YAML to some schema.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/SchemaAssociationsPreferenceInitializer.java
```diff
@@ -0,0 +1,41 @@
...
+package org.eclipse.wildwebdeveloper;
+
+import org.eclipse.core.runtime.preferences.AbstractPreferenceInitializer;
+import org.eclipse.jface.preference.IPreferenceStore;
+
+import com.google.gson.JsonObject;
+
+public class SchemaAssociationsPreferenceInitializer extends AbstractPreferenceInitializer {
+	private static final IPreferenceStore STORE = Activator.getDefault().getPreferenceStore();
+	public static final String SCHEMA_ASSOCIATIONS_PREFERENCE = "wildwebdeveloper.schema.associations";
```
mickaelistria, line 19: To prevent users from loosing previous configurations on upgrde, I suggest we keep the former preference name `wildwebdeveloper.yaml.schema`

mickaelistria (changes requested):
That's really good!
Please see 1 comment inline about compatibility.

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -1,7 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <?eclipse version="3.4"?>
 <plugin>
-   <extension-point id="org.eclipse.wildwebdeveloper.json.schema" name="Bind JSON Schema to filename pattern" schema="schema/json.schema.exsd"/>
```
mickaelistria, line 4: I see this got removed but there is a new .exsd file. Shouldn't it just be renamed?

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSonLanguageServer.java
```diff
@@ -25,29 +25,55 @@
...
+import org.eclipse.wildwebdeveloper.SchemaAssociationsPreferenceInitializer;
 import org.eclipse.wildwebdeveloper.embedder.node.NodeJSManager;
 
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+
+@SuppressWarnings("restriction")
 public class JSonLanguageServer extends ProcessStreamConnectionProvider {
 
-	private final static String SCHEMA_EXT = "org.eclipse.wildwebdeveloper.json.schema"; //$NON-NLS-1$
```
mickaelistria, line 42: Now this extension point is here, we can't remove it too quickly as it would break some adopters. The code that consumes it needs to remain, and be merged with the new code reading preferences. We can then consider deprecating the extension point later.

mickaelistria (changes requested):

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/SchemaAssociationsPreferenceInitializer.java
```diff
@@ -0,0 +1,41 @@
...
+package org.eclipse.wildwebdeveloper;
+
+import org.eclipse.core.runtime.preferences.AbstractPreferenceInitializer;
+import org.eclipse.jface.preference.IPreferenceStore;
+
+import com.google.gson.JsonObject;
+
+public class SchemaAssociationsPreferenceInitializer extends AbstractPreferenceInitializer {
+	private static final IPreferenceStore STORE = Activator.getDefault().getPreferenceStore();
+	public static final String SCHEMA_ASSOCIATIONS_PREFERENCE = "wildwebdeveloper.yaml.schema";
```
mickaelistria, line 19: That may be the right place where to also read the legacy extension point for JSon.

org.eclipse.wildwebdeveloper/plugin.xml
ndoschek: I did not add any new .exsd file, but as we need to keep the extension point, i will re-add the orginial .exsd file again.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSonLanguageServer.java
ndoschek: Ah I see, that's a fair point!
As the extension point offers patterns and no contentTypes, is it okay to pass the associations from the extension point again directly to the `JSonLanguageServer`?
I think it might be a bit misleading to include them in the preference table with its different pattern / url combination.
I would then add additional info line on the preference page, to inform the user that the schema associations from the extension points are not included but still applying.
What do you think?

mickaelistria: you're right, sticking to JSonLanguageServer seems more correct and less confusing.

mickaelistria:
Thanks a lot @ndoschek for this great contribution, and for being persistent despite my many queries!

mickaelistria:
@ndoschek Can you please add a note about it in the CHANGELOG file?

ndoschek:
@mickaelistria Yes sure, I added it to the section of the RELEASE_NOTES.md file for now, as I was kind of struggling to find a CHANGELOG file. If thats fine, then see PR #569.

Issue #315, opened 2019-12-25
log file is written on each poll


nimo23:
I get the following logs in the eclipse console on each wwd-poll and it writes a log file for each poll in the eclipse/.metadata-folder. I did not configure anything with wwd. I only installed it and it makes such a noise by making polls and repeated logs:

```
[Info]	Using @angular/language-service v9.0.0-next.8 from /Applications/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_0.5.3.201912031954/language-servers/node_modules/@angular/language-service
[Warning]	Using less efficient polling watcher. Set TSC_NONPOLLING_WATCHER to true.
[Info]	Disabling language service for /Users/workspace/ide/test/src/main/webapp/WEB-INF/tsconfig.json because it is not an Angular project.
[Error]	No config file for /Users/workspace/ide/test/src/main/webapp/WEB-INF/node_modules/crypto-browserify/example/index.html
[Info]	Using typescript v3.5.3 from /Applications/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_0.5.3.201912031954/language-servers/node_modules/typescript
[Info]	Using @angular/language-service v9.0.0-next.8 from /Applications/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_0.5.3.201912031954/language-servers/node_modules/@angular/language-service
[Info]	Log file: /Users/workspace/ide/.metadata/angular-language-server-1577301845397.log
[Info]	Angular Language Service is running under DEBUG mode
[Warning]	Using less efficient polling watcher. Set TSC_NONPOLLING_WATCHER to true.
[Info]	Using typescript v3.5.3 from /Applications/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_0.5.3.201912031954/language-servers/node_modules/typescript
[Info]	Using @angular/language-service v9.0.0-next.8 from /Applications/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_0.5.3.201912031954/language-servers/node_modules/@angular/language-service
[Info]	Log file: /Users/workspace/ide/.metadata/angular-language-server-1577301871030.log
[Info]	Angular Language Service is running under DEBUG mode
```

[...]

PR #332, opened 2020-01-25
Bug #315 - Use log file for angular LS if logToFile is enabled

jabby:
Signed-off-by: Gautier de Saint Martin Lacaze <gautier.desaintmartinlacaze@gmail.com>

PR #333, opened 2020-01-25
Bug #315 - Force use of TypeScript non-polling watchers.

jabby:


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/angular/AngularLanguageServer.java
```diff
@@ -50,7 +50,7 @@ public AngularLanguageServer() {
 
 	@Override protected ProcessBuilder createProcessBuilder() {
 		ProcessBuilder builder = super.createProcessBuilder();
-		builder.environment().put("NG_DEBUG", Boolean.toString(true));
```
mickaelistria: We should keep the NG_DEBUG set for cases when the preference to log and Angular LS is true.

jabby: I revert this modification here and I will update #332

PR #346, opened 2020-02-11
Fix browser paths & DAB ID's

### LINKED ISSUE #343: Invalid id for NodeAttachDebugDelegate
The ID is valid as an internal constant as `org.eclipse.wildwebdeveloper.launchConfiguration.nodeAttach` (see https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/node/NodeAttachDebugDelegate.java#L40)
However, the plugin.xml uses `org.eclipse.wildwebdeveloper.launchConfiguration.nodeDebugAttach` (see https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/plugin.xml#L483)
Also for consumers wanted to create such config without UI, would be fine to have it as public

AObuchow:
Fix #343


mickaelistria:
Thanks Andrew!

PR #349, opened 2020-02-11
Add option to select Chromium or Chrome browser

### LINKED ISSUE #347: Chrome debugging not possible if chromium is installed
With the current implementation of [`findChromeLocation()`](https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java#L112), Chromium will always be used if it is installed on the users machine.

I suggest a combo box be added to the Chrome debug launch configuration UI, which would allow the user to select whether they want to use Chrome or Chromium (if either is installed). This can then be used to determine the return value of  `findChromeLocation()`.

### LINKED ISSUE #348: Chromium hardcoded into Running Chrome instance debug configuration
Currently, the value of Chrome DAP `runtimeExecutable` is [hardcoded to `/usr/bin/chromium-browser`](https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachDebugDelegate.java#L41). It should call [`ChromeRunDAPDebugDelegate.findChromeLocation()`](https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java#L112).

As a result of this bug, it's impossible to attach to a running Chrome instance (only Chromium is possible).

AObuchow:
Make RunChromeDebugTab call super.InitializeFrom()

Fix #347
Fix #348

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachTab.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachTab.java
@@ -1,14 +1,11 @@
 /*******************************************************************************
- * Copyright (c) 2019 Red Hat Inc. and others.
+ * Copyright (c) 2020 Red Hat Inc. and others.
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
  * which is available at https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
- *
- * Contributors:
- *   Andrew Obuchowicz (Red Hat Inc.) 
  *******************************************************************************/
 package org.eclipse.wildwebdeveloper.debug.chrome;
 
@@ -17,25 +14,24 @@ import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.SelectionListener;
-import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Combo;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.wildwebdeveloper.Activator;
-import org.eclipse.wildwebdeveloper.debug.AbstractRunHTMLDebugTab;
+import org.eclipse.wildwebdeveloper.debug.node.AttachTab;
 
-public class RunChromeDebugTab extends AbstractRunHTMLDebugTab {
-	private Button verboseConsoleOutput;
+public class ChromeAttachTab extends AttachTab {
+
 	private Combo browserToUse;
-	
-	public RunChromeDebugTab() {
-		super.shortcut = new ChromeRunDebugLaunchShortcut();
+
+	public ChromeAttachTab() {
+		super(9222);
 	}
-	
+
 	@Override
 	public void createControl(Composite parent) {
 		super.createControl(parent);
-		new Label(resComposite, SWT.NONE).setText("Run with: ");
+		new Label(resComposite, SWT.NONE).setText(Messages.ChromeAttachTab_runWith);
 		browserToUse = new Combo(resComposite, SWT.VERTICAL | SWT.DROP_DOWN | SWT.BORDER | SWT.READ_ONLY);
 		browserToUse.add(ChromeRunDAPDebugDelegate.CHROMIUM);
 		browserToUse.add(ChromeRunDAPDebugDelegate.CHROME);
@@ -43,15 +39,13 @@ public class RunChromeDebugTab extends AbstractRunHTMLDebugTab {
 			setDirty(true);
 			updateLaunchConfigurationDialog();
 		}));
-		Composite emptySpace = new Composite(resComposite, SWT.NONE);
-		verboseConsoleOutput = new Button(resComposite, SWT.CHECK);
-		verboseConsoleOutput.setText("Verbose console output");
-		verboseConsoleOutput.addSelectionListener(SelectionListener.widgetSelectedAdapter((e) -> {
-			setDirty(true);
-			updateLaunchConfigurationDialog();
-		}));
 	}
-	
+
+	@Override
+	public void setDefaults(ILaunchConfigurationWorkingCopy configuration) {
+		super.setDefaults(configuration);
+	}
+
 	@Override
 	public void initializeFrom(ILaunchConfiguration configuration) {
 		super.initializeFrom(configuration);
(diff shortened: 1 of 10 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachTab.java
```diff
@@ -0,0 +1,57 @@
...
+		browserToUse = new Combo(resComposite, SWT.VERTICAL | SWT.DROP_DOWN | SWT.BORDER | SWT.READ_ONLY);
+		browserToUse.add(ChromeRunDAPDebugDelegate.CHROMIUM);
+		browserToUse.add(ChromeRunDAPDebugDelegate.CHROME);
+		browserToUse.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
+			setDirty(true);
+			updateLaunchConfigurationDialog();
+		}));
+	}
+
+
```
AObuchow: remove empty line

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachTab.java
```diff
@@ -0,0 +1,57 @@
...
+	private Combo browserToUse;
+
+	public ChromeAttachTab(int defaultPort) {
+		super(defaultPort);
+	}
+
+	@Override
+	public void createControl(Composite parent) {
+		super.createControl(parent);
+		new Label(resComposite, SWT.NONE).setText("Run with: ");
```
AObuchow: Externalize

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
```diff
@@ -106,11 +108,24 @@ static File findDebugAdapter() {
...
+	static String findChromeLocation(ILaunchConfiguration configuration) {
+		String res = null;
+		try {
+			if (configuration.getAttribute(RUNTIME_EXECUTABLE, "no runtime executable set").equals(CHROMIUM)) {
+				res = InitializeLaunchConfigurations.which("chromium-browser");
+			} else {
+				res = InitializeLaunchConfigurations.which("google-chrome-stable");
+			}
+		} catch (CoreException e) {
+			// TODO Auto-generated catch block
```
AObuchow: Remove, use logger

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachTab.java
```diff
@@ -0,0 +1,57 @@
+package org.eclipse.wildwebdeveloper.debug.chrome;
```
mickaelistria: Missing copyright header.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
```diff
@@ -106,11 +108,24 @@ static File findDebugAdapter() {
...
 		return null;
-		
 	}
 
-	private String findChromeLocation() {
-		String res = InitializeLaunchConfigurations.which("chromium-browser");
+	static String findChromeLocation(ILaunchConfiguration configuration) {
+		String res = null;
+		try {
+			if (configuration.getAttribute(RUNTIME_EXECUTABLE, "no runtime executable set").equals(CHROMIUM)) {
```
mickaelistria: In this case, let the default value be `chromium-browser` so we save a null check after that.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
```diff
@@ -43,8 +43,10 @@
 
 	static final String VERBOSE = "verbose";
 	private static final String TRACE = "trace";
-	private static final String RUNTIME_EXECUTABLE = "runtimeExecutable";
+	public static final String RUNTIME_EXECUTABLE = "runtimeExecutable";
 	private static final String SOURCE_MAPS = "sourceMaps";
+	public static final String CHROMIUM = "Chromium";
+	public static final String CHROME = "Chrome";
```
mickaelistria: Should those value be the actual `chromium-browser` vs `google-chrome-stable` ?
And the attribute could be a free text (or an editable combo) so people can pass another variation of the runtimeExecutable (can be for instance another path to a local Chromium build).

AObuchow:
> I think this could be more open to alternative names/locations for the runtimeExecutable, so it would allow more customization. For example, I'm not sure those names are the same on all OS, so I think we may need more flexibility than a restricted choice here.
> I think this should be an open combo box, with the values `chromium-browser` and `google-chrome-stable` pre-set in case those are resolved to actual path.

You raise a good point and I totally agree.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
AObuchow: Good call

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
AObuchow: Yes I totally agree on both those points.
Months ago I wanted the user to be able to set the location of their Firefox executable, so this could encompass that feature for both Firefox and Chrome.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachTab.java
```diff
@@ -0,0 +1,64 @@
...
+		browserToUse.add(ChromeRunDAPDebugDelegate.CHROMIUM);
+		browserToUse.add(ChromeRunDAPDebugDelegate.CHROME);
+		browserToUse.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
+			setDirty(true);
+			updateLaunchConfigurationDialog();
+		}));
+	}
+
+	@Override
+	public void setDefaults(ILaunchConfigurationWorkingCopy configuration) {
```
mickaelistria: In case there is nothing different from super, let's just remove the method.

mickaelistria:
Rebased and merged.

mickaelistria:
Rebased and merged.

PR #355, opened 2020-02-13
Upgrade VSCode LS

mickaelistria:
+ Add formatter initializeOptions to JSON LS

### DIFF
--- a/org.eclipse.wildwebdeveloper/pom.xml
+++ b/org.eclipse.wildwebdeveloper/pom.xml
@@ -38,7 +38,7 @@
 							<goal>wget</goal>
 						</goals>
 						<configuration>
-							<url>https://az764295.vo.msecnd.net/stable/f06011ac164ae4dc8e753a3fe7f9549844d15e35/code-stable-1565886317.tar.gz</url>
+							<url>https://az764295.vo.msecnd.net/stable/ae08d5460b5a45169385ff3fd44208f431992451/code-stable-1580986866.tar.gz</url>
 							<unpack>true</unpack>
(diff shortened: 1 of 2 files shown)


org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -114,14 +115,14 @@
...
-								<arg>./node-debug2-1.33.0.tgz</arg>
+								<arg>./node-debug2-1.42.1.tgz</arg>
 								<arg>typescript@3.7.2</arg>
 								<arg>typescript-language-server@0.4.0</arg>
 								<arg>typescript-lit-html-plugin@0.9.0</arg>
 								<arg>typescript-plugin-css-modules@2.1.0</arg>
 								<arg>./vscode-css-languageserver-1.0.0.tgz</arg>
 								<arg>./vscode-html-languageserver-1.0.0.tgz</arg>
-								<arg>vscode-json-languageserver@1.2.2</arg>
+								<arg>./vscode-json-languageserver-1.2.2.tgz</arg>
```
akurtakov: So the json ls on npm.js is different from the one in vscode?

mickaelistria: yes, and the one on npmjs is currently bugged: https://github.com/microsoft/vscode/issues/90634
Actually, I think it's often best ot just pick what's directly in VSCode rather than on npm; it seems like it's better tested.

mickaelistria:
Opened CQ http://dev.eclipse.org/ipzilla/show_bug.cgi?id=21648

PR #368, opened 2020-02-26
Add TypeScript website debug instructions to FAQ

AObuchow:
Enable sourceMaps in Chrome Launch Debug Delegate

### DIFF
--- a/TIPS_and_FAQ.md
+++ b/TIPS_and_FAQ.md
@@ -15,6 +15,24 @@
   c. select the interesting typescript source folder in "Select working set of relevant resources" to include the source folder.
 5. in Refresh, select what needs to be refreshed upon build.
 
+#### Debug a website developed in TypeScript?
+
+1. Ensure the `sourceMap` property of your `tsconfig.json` is set to `true`.
+Example `tsconfig.json`:
+```JSON
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "outDir": "out",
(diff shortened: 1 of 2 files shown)


AObuchow:
Note: this only shows how to debug TypeScript with Chrome (I haven't figuredo out how to get it working in Firefox yet, it might require more work for Firefox). It also assumes that the `out` folder lives in the project folder. A future patch should be made to allow modification of the sourceMaps path.

AObuchow:
If anyone wants to verify this works, https://code.visualstudio.com/docs/typescript/typescript-debugging provides an examplary project to test it out. (It works for me)

jabby:
I will try this tomorrow.

mickaelistria:
That looks very good and I trust this works, so let's merge it now, and improve it if necessary in the future.
It would be interesting to also document the case of a plain "microservice" written in typescript (ie without an html file as entrypoint).

AObuchow:
> I will try this tomorrow.

Thanks @jabby, despite the fact this was merged it would still be greatly appreciated if you could test it out to ensure it's valid.

AObuchow:
> It would be interesting to also document the case of a plain "microservice" written in typescript (ie without an html file as entrypoint).

Yes this is something I've wanted to do for a while as I think microservice debuging/development is essential for web developers in 2020.

I'm almost certain it should be possible (I had it working on a local server which used webpack + source maps), although a modification will need to be made to the launch configuration to accept a URL instead of just files.

PR #380, opened 2020-03-14
Add documentation for importing existing projects

AugustNagro:


README.md
```diff
@@ -85,6 +85,25 @@ In an Eclipse target platform definition:
 
 Wild Web Developer is based on the Eclipse Generic Editor framework from [Eclipse Platform](https://projects.eclipse.org/projects/eclipse.platform), [Eclipse LSP4E](https://projects.eclipse.org/projects/technology.lsp4e) and [Eclipse TM4E](https://github.com/eclipse/tm4e) in order to provide editors based on TextMate grammars, Visual Studio Code HTML CSS & JSON Language Servers, Theia's [TypeScript and JavaScript language servers](https://github.com/theia-ide/typescript-language-server), Red Hat's [xml-languageserver](https://github.com/angelozerr/lsp4xml) and [yaml-language-server](https://github.com/redhat-developer/yaml-language-server/)...; and on Eclipse Debug stack, LSP4E Debug Adapter Protocol support and Visual Studio Code Node Debug Adapter... to provide debugging.
 
+## Importing Existing Projects
```
mickaelistria: The recommended way would just be to use
* either File > Open Projects from Filesystem..., or
* File > Import > General > Projects from Folder or Archive

Both do lead to a similar result as what you suggest, but are much simpler to use.

Also, for this task, it would be great if the documentation could just link to existing documetation from help.eclipse.org instead of duplicating some similar content.

AugustNagro:
Yes you're right.. I'm newer to eclipse so I missed these much nicer ways.

Closing this request.

PR #385, opened 2020-03-18
Add support for web app debugging

AObuchow:


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
```diff
@@ -100,7 +95,30 @@ public void createControl(Composite parent) {
...
+		urlText = new Text(resComposite, SWT.BORDER);
+		GridData urlTextGD = new GridData(SWT.FILL, SWT.DEFAULT, true, false);
+		urlTextGD.horizontalSpan = 2;
+		urlText.setLayoutData(urlTextGD);
+		urlText.addModifyListener(e -> {
+			setDirty(true);
+			updateLaunchConfigurationDialog();
+		});
+		// Select file debugging by default
+		fileRadio.setSelection(true);
```
AObuchow: This creates an annoying bug, which is that the File radio button is always checked, even if the URL radio button was checked previously.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
```diff
@@ -130,10 +148,40 @@ public void createControl(Composite parent) {
 				updateLaunchConfigurationDialog();
 			}
 		}));
+
```
AObuchow: Remove whitespace

TIPS_and_FAQ.md
```diff
@@ -33,6 +33,17 @@ Example `tsconfig.json`:
...
+
+1. Start your web app's Node.JS web server (usually done through a NPM script in your package.json, eg. `npm start` ). This can be done through the commandline or by right-clicking on a package.json in the Project Explorer => Run as => NPM...
+2. Once your web app is running, take note of the local URL it's running on. For React project's, it's by default `http://localhost:3000/`
+3. Right click on your project's root directory in the Project Explorer => Debug as => Debug configurations...
+4. Create a launch or attach debug configuration for Chrome or Firefox, depending on your preference.
+5. For launch debug configurations, select the URL radio button and enter the URL from step 2. For attach debug configurations, simply enter the URL from step 2.
+6. Ensure the working directory is correctly set to your project's root folder.
+7. Click `apply` followed by `debug` in the bottom right corner of the debug configuration
+
+
```
AObuchow: Remove whitespace

org.eclipse.wildwebdeveloper.xml/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: XML support using Language Server
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper.xml;singleton:=true
-Bundle-Version: 0.6.1.qualifier
+Bundle-Version: 0.6.2.qualifier
```
mickaelistria: You shouldn't have to bump version here, as the bundle payload didn't change.

org.eclipse.wildwebdeveloper.xml/pom.xml
```diff
@@ -7,7 +7,7 @@
 		<version>0.5.0-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-plugin</packaging>
-	<version>0.6.1-SNAPSHOT</version>
+	<version>0.6.2-SNAPSHOT</version>
```
mickaelistria: You shouldn't have to bump version here, as the bundle payload didn't change.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
mickaelistria: Radio selection should be updated in the `initializeFrom()` method then.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachDebugDelegate.java
```diff
@@ -36,8 +36,12 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
 		Map<String, Object> param = new HashMap<>();
 		param.put(ADDRESS, configuration.getAttribute(ADDRESS, "no address defined")); //$NON-NLS-1$
 		param.put(AbstractHTMLDebugDelegate.PORT, configuration.getAttribute(NodeAttachDebugDelegate.PORT, 9229));
-		param.put("url", "https://www.google.ca/");
-		param.put("runtimeExecutable", "/usr/bin/chromium-browser");
+		String url = configuration.getAttribute(ChromeRunDAPDebugDelegate.URL, "");
+		if (!url.equals("")) {
```
mickaelistria: `!url.isEmpty()`

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxRunDebugLaunchShortcut.java
```diff
@@ -12,12 +12,22 @@
...
 public class FirefoxRunDebugLaunchShortcut extends AbstractHTMLDebugAdapterLaunchShortcut {
 
 	public FirefoxRunDebugLaunchShortcut() {
 		super(FirefoxRunDABDebugDelegate.ID);
 	}
+	
+	@Override
+	public void configureLaunchConfiguration(File file, ILaunchConfigurationWorkingCopy wc) {
+		super.configureLaunchConfiguration(file, wc);
+		wc.setAttribute(ChromeRunDAPDebugDelegate.URL, "");
```
mickaelistria: Why even setting it here?

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxRunDABDebugDelegate.java
```diff
@@ -58,7 +61,27 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
...
-		param.put(FILE, configuration.getAttribute(AbstractHTMLDebugDelegate.PROGRAM, "No program path set").trim()); //$NON-NLS-1$
+
+		// File or URL to debug 
+		String url = configuration.getAttribute(ChromeRunDAPDebugDelegate.URL, "");
+		if (!url.equals("")) {
+			param.put(ChromeRunDAPDebugDelegate.URL, url);
+			File projectDirectory = new File(configuration.getAttribute(AbstractHTMLDebugDelegate.CWD, ""));
+			IProject project = ResourcesPlugin.getWorkspace().getRoot().getProject(projectDirectory.getName());
+			try {
+				param.put("webRoot", project.getRawLocation().toString() + "/");
```
mickaelistria: why nt just `param.put("webRoot", projectDirectory.getAbsolutePath())` ?

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/RunFirefoxDebugTab.java
```diff
@@ -30,12 +30,11 @@ public RunFirefoxDebugTab() {
 	public void createControl(Composite parent) {
 		super.createControl(parent);
 		reloadOnChange = new Button(resComposite, SWT.CHECK);
-		reloadOnChange.setText("Reload on change");
+		reloadOnChange.setText("Reload on change"); //$NON-NLS-1$
```
mickaelistria: This will need to be internationalized, so let's extract it to some `Messages` class and *not* mark it as `//$NON-NLS-1`. NON-NLS means that a string shouldn't be translated, this one should.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/node/AttachTab.java
```diff
@@ -51,20 +65,47 @@ public void createControl(Composite parent) {
 			setDirty(true);
 			updateLaunchConfigurationDialog();
 		});
+		
+		new Label(resComposite, SWT.NONE).setText(Messages.RunProgramTab_workingDirectory);
+		Composite workingDirectoryComposite = new Composite(resComposite, SWT.NONE);
```
mickaelistria: In which case do we need it to be set? So far, it's been working well without it.

org.eclipse.wildwebdeveloper.xml/META-INF/MANIFEST.MF
AObuchow: Sounds good, I think this is just a leftover from earlier work or something. It wasn't letting me compile with maven for some reason without a bump. Will fix.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachDebugDelegate.java
AObuchow: Thanks, will use both of those.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/FirefoxRunDABDebugDelegate.java
AObuchow: Thanks, much better.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/firefox/RunFirefoxDebugTab.java
AObuchow: Ok, will externalize.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/node/AttachTab.java
AObuchow: Only for web app debuging in Firefox and Chrome, not Node.js. I'll move it into the child classes.

org.eclipse.wildwebdeveloper.tests/.project
```diff
@@ -20,8 +20,14 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
+		<buildCommand>
```
AObuchow: This commit was accidental. 
I can do a git cherry pick when this patch is OK for a merge to get rid of this useless commit.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
```diff
@@ -76,11 +75,20 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
 			param.put(AbstractHTMLDebugDelegate.ENV, envJson);
 		}
 		
+		// File or URL to debug 
+		String url = configuration.getAttribute(URL, "");
+		if (!url.equals("")) {
+			param.put(URL, url);
+		} else {
+			param.put("file", configuration.getAttribute(AbstractHTMLDebugDelegate.PROGRAM, "no program path defined")); //$NON-NLS-1$
+
```
AObuchow: Remove whitespace

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
```diff
@@ -76,11 +75,20 @@ public void launch(ILaunchConfiguration configuration, String mode, ILaunch laun
...
+			param.put("file", configuration.getAttribute(AbstractHTMLDebugDelegate.PROGRAM, "no program path defined")); //$NON-NLS-1$
+			
+		}
+		
 		// Chrome working directory
 		String cwd = configuration.getAttribute(AbstractHTMLDebugDelegate.CWD, "").trim(); //$NON-NLS-1$
 		if (!cwd.isEmpty()) {
 			param.put(AbstractHTMLDebugDelegate.CWD, cwd);
-		}
+		}
```
AObuchow: remove whitespace

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDebugLaunchShortcut.java
```diff
@@ -27,6 +27,8 @@ public ChromeRunDebugLaunchShortcut() {
 	public void configureLaunchConfiguration(File file, ILaunchConfigurationWorkingCopy wc) {
 		super.configureLaunchConfiguration(file, wc);
 		wc.setAttribute(ChromeRunDAPDebugDelegate.RUNTIME_EXECUTABLE, ChromeRunDAPDebugDelegate.CHROMIUM);
+		wc.setAttribute(ChromeRunDAPDebugDelegate.URL, "");
+
```
AObuchow: remove whitespace

org.eclipse.wildwebdeveloper.xml/.project
```diff
@@ -20,8 +20,14 @@
...
 			</arguments>
 		</buildCommand>
+		<buildCommand>
+			<name>org.eclipse.m2e.core.maven2Builder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
 	</buildSpec>
 	<natures>
+		<nature>org.eclipse.m2e.core.maven2Nature</nature>
```
mickaelistria: Please amend to remove it to keep the commit as minimal as possible

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
```diff
@@ -148,6 +190,18 @@ public void initializeFrom(ILaunchConfiguration configuration) {
 			this.argumentsText.setText(configuration.getAttribute(AbstractHTMLDebugDelegate.ARGUMENTS, "")); //$NON-NLS-1$
 			this.workingDirectoryText.setText(
 					configuration.getAttribute(AbstractHTMLDebugDelegate.CWD, pathOrEmpty(getSelectedProject())));
+			this.urlText.setText(configuration.getAttribute(ChromeRunDAPDebugDelegate.URL, "")); //$NON-NLS-1$
+			boolean urlDebugMode = configuration.getAttribute(AbstractHTMLDebugDelegate.URL_DEBUG, false);
```
mickaelistria: I don't think it's worth storing the file if URL is used or vice-versa.
IMO, we can get rid of the URL_DEBUG attribute and just check `file == null` to know whether it's a URL or a File Launch.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
```diff
@@ -156,13 +210,18 @@ public void initializeFrom(ILaunchConfiguration configuration) {
...
+			configuration.setAttribute(ChromeRunDAPDebugDelegate.URL, urlText.getText());
+		}
+
 		configuration.setAttribute(AbstractHTMLDebugDelegate.ARGUMENTS, this.argumentsText.getText());
 		String workingDirectory = this.workingDirectoryText.getText();
 		configuration.setAttribute(AbstractHTMLDebugDelegate.CWD, workingDirectory);
 		configuration.setAttribute(DebugPlugin.ATTR_WORKING_DIRECTORY, workingDirectory);
 		configuration.setMappedResources(ResourcesPlugin.getWorkspace().getRoot().findFilesForLocationURI(new File(programPath).toURI()));
-	
+		configuration.setAttribute(AbstractHTMLDebugDelegate.URL_DEBUG, urlRadio.getSelection());
```
mickaelistria: Same as above; I don't think we need to store that as the configuration shouldn't keep URL in case File is used, nor File when URL is used.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
```diff
@@ -156,13 +210,18 @@ public void initializeFrom(ILaunchConfiguration configuration) {
 	@Override
 	public void performApply(ILaunchConfigurationWorkingCopy configuration) {
 		String programPath = this.programPathText.getText();
-		configuration.setAttribute(AbstractHTMLDebugDelegate.PROGRAM, programPath);
+		if (programPathText.isEnabled()) {
+			configuration.setAttribute(AbstractHTMLDebugDelegate.PROGRAM, programPath);
```
mickaelistria: this should probably set the URL attribute to null

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
```diff
@@ -156,13 +210,18 @@ public void initializeFrom(ILaunchConfiguration configuration) {
...
+		} else if (urlText.isEnabled()) {
+			configuration.setAttribute(ChromeRunDAPDebugDelegate.URL, urlText.getText());
```
mickaelistria: This should probably set the File attribute to null.

AObuchow:
> Is `workingDirectory` useful when using File ? If not, then the related field and widgets should also be disabled when File is selected.

Working directory is useful when using File, as we can watch a working directory for changes and reload on change (I know Firefox can do this, not sure about Chrome).

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
```diff
@@ -148,6 +190,17 @@ public void initializeFrom(ILaunchConfiguration configuration) {
 			this.argumentsText.setText(configuration.getAttribute(AbstractHTMLDebugDelegate.ARGUMENTS, "")); //$NON-NLS-1$
 			this.workingDirectoryText.setText(
 					configuration.getAttribute(AbstractHTMLDebugDelegate.CWD, pathOrEmpty(getSelectedProject())));
+			this.urlText.setText(configuration.getAttribute(ChromeRunDAPDebugDelegate.URL, "")); //$NON-NLS-1$
+			if (urlText.getText().isEmpty()) {
```
AObuchow: I ended up just checking if the URL text is empty as opposed to the File parameter so that by default, the File radio button will be selected. 

Also, I wasn't able to do `configuration.getAttribute(ChromeRunDAPDebugDelegate.URL, null)`, it seemed to result in a compile error (or at least, Eclipse didn't like it). However, this works in my testing.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/AbstractRunHTMLDebugTab.java
AObuchow: I set it to an empty string `""` since setting it to `null` resulted in the compile error `The method setAttribute(String, String) is ambiguous for the type ILaunchConfigurationWorkingCopy`

mickaelistria:
Can you please re-explain ho can "URL" on Attach configuration be useful? What user-story does it cover?

AObuchow:
> Can you please re-explain ho can "URL" on Attach configuration be useful? What user-story does it cover?

I'm not sure if it's entirely useful. The user story would be when the user already has a browser open with debugging mode enabled, and they want to debug their web app in the existing browser instance. 

Personally, when I start up a react web app, a Firefox instance is automatically opened at the URL it's being hosted on. Perhaps if my configuration was tweaked, a Firefox instance with the remote debug flag would be launched after starting the server, which could then be attached to.

In all honesty, I'm OK with not having URL debugging for "attach" debug configurations. We could leave it out until someone makes a request for it.

mickaelistria:
> In all honesty, I'm OK with not having URL debugging for "attach" debug configurations. 

OK, please remove it then.
The "attach" mode means that the user debug workflow is "driven" by usage of a pre-existing browser. So we can assume that the user would open URL in the browser more than they would configure it from the IDE.

AObuchow:
> > In all honesty, I'm OK with not having URL debugging for "attach" debug configurations.
> 
> OK, please remove it then.
> The "attach" mode means that the user debug workflow is "driven" by usage of a pre-existing browser. So we can assume that the user would open URL in the browser more than they would configure it from the IDE.
 
Ok I'll give it a try and see if you can still debug a web app like that. IIRC you *need* to give the "url" parameter to debug a web app with firefox as mentioned [here](https://github.com/firefox-devtools/vscode-firefox-debug#attach), specifically 

> If your application is running on a Webserver, you need to add the url and webRoot properties to the configuration (as in the second launch configuration example above).

However, like I said, it might not be that useful to debug a web app on a already-running browser instance. So if removing the url parameter prevents web app debuging, I'm ok with that.

AObuchow:
> Ok I'll give it a try and see if you can still debug a web app like that. IIRC you _need_ to give the "url" parameter to debug a web app with firefox as mentioned [here](https://github.com/firefox-devtools/vscode-firefox-debug#attach)

It seems for Attach debugging with Firefox, if you pass the webRoot (who's value is the working directory), you must pass a URL. However, it works even if the url given is `""` (an empty string).
So I can safely remove the URL text field in the UI of the attach configurations without preventing web app debugging with attach debug configurations.

AObuchow:
> It seems for Attach debugging with Firefox, if you pass the webRoot (who's value is the working directory), you must pass a URL. However, it works even if the url given is `""` (an empty string).
> So I can safely remove the URL text field in the UI of the attach configurations without preventing web app debugging with attach debug configurations.

Nevermind, I made a mistake in my testing...
- If you pass the `url` parameter to the Firefox debugger, then it *must* not be an empty string, and you *must* pass the `webRoot` parameter.
- If you pass the `webRoot` parameter, then you must pass the `url` parameter.

However, if you omit both and have a React application running, you can still hit breakpoints...
I'm testing with a react application written in JavaScript, not TypeScript. So maybe it's a bad test.

AObuchow:
Further investigation with Chrome shows that:
- If no `url` parameter is passed, or if it's an empty string, attach configurations will uselessly launch a new chrome instance, and attach to the previously-already-running instance of chrome
- If the `url` parameter is passed, but it's not the `url` of the web app (eg. `https://youtube.com` is given), it will attach to the running instance (without creating another useless instance of Chrome) and open a new tab at the url which was given.

In both instances, breakpoints in my react app were still hit. 

I'm going to see if this occurs in VSCode and if it does, report it upstream.

mickaelistria:
I'm a bit confused, what do you conclude from all that? Should we keep URL for Attach or can we get rid of it and achieve the same result without additional complexity?

AObuchow:
@mickaelistria sorry for so many notes, I was writing those for my future self primarily hahah.

My conclusion as that for Firefox attach, we can get rid of URL and working directory fields.
For Chrome, it's best to keep the URL field as Chrome will open the URL given by the user in a new tab.

[We were already giving "google.ca" as the URL for chrome attach to work](https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeAttachDebugDelegate.java#L39), which might have confused users as to why google was suddenly being opened in a new tab. Thus letting them set the site to open on attach is a better solution than forcing a default URL.

Additionally, if we were to pass no URL at all (not even google.ca) then a new Chrome instance will be launched, which I think is a bad UX.

mickaelistria:
OK, let's do this.

AObuchow:
@mickaelistria okay great sounds good, the latest commit I pushed should be good for a merge.

AObuchow:
Correction: now it should be good. I made it resort to setting the `url` to google.ca if no url is set by the user, for the chrome attach debug configuration

mickaelistria:
Merged as dbe237fe00ce116fba8a09309a0f77f9e2e2820b . Thanks!

PR #394, opened 2020-03-27
DRAFT: A Simple JSP editor

vrubezhny:
Provides the editing facility for *.jsp", "*.jspf" and "*.tag" files,
Provides:
- For HTML content type: syntax highlighting, content assist and error reporting (via 'HTMLLanguageServer')
- For Java content type: syntax highlighting only

### DIFF
--- a/org.eclipse.wildwebdeveloper/plugin.xml
+++ b/org.eclipse.wildwebdeveloper/plugin.xml
@@ -389,6 +389,61 @@
      <icon contentType="org.eclipse.wildwebdeveloper.ts" icon="icons/tsEditorIcon.png"/>
    </extension>
 
+   <!-- Java Server Pages (JSP) -->   
+   <extension
+         point="org.eclipse.core.contenttype.contentTypes">
+      <content-type
+            base-type="org.eclipse.wildwebdeveloper.parent"
+            file-extensions="jsp,jspf,tag"
+            id="org.eclipse.wildwebdeveloper.jsp"
+            name="JavaServerPages"
+            priority="normal">
+      </content-type>
+   </extension>
+
+   <extension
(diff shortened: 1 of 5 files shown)


org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -389,6 +389,61 @@
      <icon contentType="org.eclipse.wildwebdeveloper.ts" icon="icons/tsEditorIcon.png"/>
    </extension>
 
+   <!-- Java Server Pages (JSP) -->   
+   <extension
+         point="org.eclipse.core.contenttype.contentTypes">
+      <content-type
+            base-type="org.eclipse.wildwebdeveloper.parent"
```
mickaelistria, line 396: I think you can declare that base type is `org.eclipse.wildwebdeveloper.xml`...

org.eclipse.wildwebdeveloper/plugin.xml
```diff
@@ -389,6 +389,61 @@
...
+            file-extensions="jsp,jspf,tag"
+            id="org.eclipse.wildwebdeveloper.jsp"
+            name="JavaServerPages"
+            priority="normal">
+      </content-type>
+   </extension>
+
+   <extension
```
mickaelistria, line 404: ....and get rid of the language server redefinition as the HTML language server would be used (because it's mapped to a parent type)

mickaelistria:
Thanks for the interesting experiment which IMO validates some minimal support can be affordable. But we won't include this in Wild Web Developer.
If some JSP support via Wild Web Developer is to be provided; it would better fit in Eclipse WTP/JST or JBoss Tools for instance (ping @nitind @jeffmaury for awareness if one of you is willing to continue this work).

Issue #396, opened 2020-03-28
Eclipse Warning: Could not find node.js for FreeBSD


daemonblade:
I'm running Eclipse 2020-03 on FreeBSD, and have just installed Wild Web Developer 0.9.0.202003192213. However, on startup on FreeBSD and on startup, it complains with "Could not find node.js...".

However, I do have node installed:
```
onyx:~,10:30am> which node
/usr/local/bin/node
onyx:~,11:13am> node --version
v13.10.1
```

I have looked at "InitializeLaunchConfigurations.java", and the which() method incorrectly assumes that "/bin/bash" exists. This is a Linuxism. Wouldn't it be better to inspect $PATH and then see if an executable file exists with the name "node"?

Adding hack in getDefaultNodePath() wouldn't quite work for FreeBSD as it's not one of the return values defined in Platform.getOS().

daemonblade:
Something like the following, perhaps:
```
public static String which(String program) {
    String paths [] = System.getenv ("PATH").split (System.getProperty ("path.separator"));
    for (String path : paths) {
        File node = new File (path, program);
        if (node.canExecute ())
            return node.getAbsolutePath ();
    }
    return null;
}
```

mickaelistria:
Do you think you can provide a pull request for that?

daemonblade:
Thanks!

mickaelistria:
Build including fix should be available in a dozen of minutes at https://download.eclipse.org/wildwebdeveloper/snapshots/

daemonblade:
FYI: Build 0.9.0.202003301134 works perfectly.

PR #397, opened 2020-03-29
Issue #396: which() to inspect PATH environment var for executables

daemonblade:
Please review, as I'm doing this blind. I don't have a dev-env for wildwebdeveloper set up.

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java
@@ -17,6 +17,7 @@
 package org.eclipse.wildwebdeveloper;
 
 import java.io.BufferedReader;
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.file.Files;
@@ -68,6 +69,14 @@ public class InitializeLaunchConfigurations {
 	}
 
 	public static String which(String program) {
+
+		String[] paths = System.getenv("PATH").split(System.getProperty("path.separator"));
+		for (String path : paths) {
+			File exe = new File(path, program);
+			if (exe.canExecute())
+				return exe.getAbsolutePath();
+		}
+
 		String res = null;
 		String[] command = new String[] { "/bin/bash", "-c", "-l", "which " + program};
 		if (Platform.getOS().equals(Platform.OS_WIN32)) {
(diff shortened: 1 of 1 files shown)


akurtakov:
@mbooth101 Mat, I would like to hear your opinion on that from flatpak/sandbox POV.

mbooth101:
Wouldn't it be better to simply change ```/bin/bash``` to ```/bin/sh``` ?

On Linux ```/bin/sh``` is a symlink to the distro's default shell anyway. On Fedora/RHEL it is a symlink to bash, on Debian it is a symlink to dash, etc.

daemonblade:
I'm puzzled as to why /bin/sh (or /bin/bash) `which` would discover something on ${PATH} that a manual inspection wouldn't.

mbooth101:
> I'm puzzled as to why /bin/sh (or /bin/bash) `which` would discover something on ${PATH} that a manual inspection wouldn't.

We use a shell because a user may have additional paths configured in their ```~/.profile``` (or ```~/.bash_profile```) that we could not check without invoking a login shell, i.e.: ```/bin/sh -l```

daemonblade:
In which case I agree that the quickest way to do this is change the shell to /bin/sh

mickaelistria:
@daemonblade Ok, so can you please amend your patch and `push --force` to your branch to update the PR. By the way, you'll need to add the "Signed-Off-By..." footer in your commit with  `git commit --amend --signoff` and ensure you signed the Eclipse Contributor Agreement at https://accounts.eclipse.org/legal/eca/

daemonblade:
@mickaelistria Reading thru' the original code, /bin/bash uses the "-l" option. Unfortunately, this option isn't present in POSIX sh.

Can I suggest that the Java based PATH inspection snippet be place before the current Runtime.exec() attempt? I've amended my patch to show you what I mean.

mickaelistria:
I'm fine with this patch as long as 1. CI tests still pass on regular Linux, 2. you confirm it fixes your issue and 3. @mbooth101 confirms it doesn't break Flatpak deliveries.

mbooth101:
> @mickaelistria Reading thru' the original code, /bin/bash uses the "-l" option. Unfortunately, this option isn't present in POSIX sh.
> 
> Can I suggest that the Java based PATH inspection snippet be place before the current Runtime.exec() attempt? I've amended my patch to show you what I mean.

Yeah, I think that's fine. When no exe is found in any PATH, then it will fall through and check with which.

mickaelistria:
So that's good for me, good for CI, good for @mbooth101 
@daemonblade Do you confirm it works for you and it's good to merge?

daemonblade:
Yes, good to merge

mickaelistria:
Thanks all, and particularly @daemonblade !

PR #401, opened 2020-04-07
Bug 559907 - Consider opening js files with GenericEditor/WildWebDeve…

vrubezhny:
…loper by default

The PR adds a workaroung that makes Generic Text Editor to be a default editor for "*.js" files on a new workspace. 

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
@@ -1,0 +1,55 @@
+/*******************************************************************************
+ * Copyright (c) 2020 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Victor Rubezhny (Red Hat Inc.) - initial implementation
+ *******************************************************************************/
+package org.eclipse.wildwebdeveloper;
+
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.content.IContentType;
+import org.eclipse.ui.IEditorDescriptor;
+import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.IEditorRegistry;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.internal.genericeditor.ExtensionBasedTextEditor;
+import org.eclipse.ui.internal.genericeditor.GenericEditorWithContentTypeIcon;
+import org.eclipse.ui.internal.genericeditor.GenericEditorWithIconAssociationOverride;
+
+public class JSEditorAssociationOverride extends GenericEditorWithIconAssociationOverride {
+	private final IEditorDescriptor genericEditorWithJSIcon;
+	private final IContentType wildWebDeveloperJSContentType;
+
+	public JSEditorAssociationOverride() {
+		IEditorRegistry editorReg = PlatformUI.getWorkbench().getEditorRegistry();
+		IEditorDescriptor genericEditorDescriptor = editorReg.findEditor(ExtensionBasedTextEditor.GENERIC_EDITOR_ID);
+		genericEditorWithJSIcon = genericEditorDescriptor != null ? new GenericEditorWithContentTypeIcon("*.js", genericEditorDescriptor) : null;
+		wildWebDeveloperJSContentType = Platform.getContentTypeManager().getContentType("org.eclipse.wildwebdeveloper.js");
+	}
+	
+	@Override
+	public IEditorDescriptor overrideDefaultEditor(IEditorInput editorInput, IContentType contentType,
+			IEditorDescriptor editorDescriptor) {
+		return editorInput != null ? overrideDefaultEditor(editorInput.getName(), contentType, editorDescriptor)
+				: editorDescriptor;
+	}
+	
+	// Workaround for https://bugs.eclipse.org/bugs/show_bug.cgi?id=559907
+	@Override
+	public IEditorDescriptor overrideDefaultEditor(String fileName, IContentType contentType,
+			IEditorDescriptor editorDescriptor) {
+		if (wildWebDeveloperJSContentType != null && genericEditorWithJSIcon != null) {
+			if(wildWebDeveloperJSContentType.equals(contentType) || 
+					wildWebDeveloperJSContentType.isAssociatedWith(fileName)) { 
+				return genericEditorWithJSIcon;
+			}
+		}
+		return super.overrideDefaultEditor(fileName, contentType, editorDescriptor);
+	}
+}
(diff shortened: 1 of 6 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// Workaround for https://bugs.eclipse.org/bugs/show_bug.cgi?id=559907
+	private IEditorDescriptor getEditorDescriptorForFile(IEditorDescriptor defaultDescriptor, String fileName) {
+		if (defaultDescriptor != null && !ExtensionBasedTextEditor.GENERIC_EDITOR_ID.equals(defaultDescriptor.getId())
+				&& fileName != null && !fileName.isEmpty()) {
+			int index = fileName.lastIndexOf('.');
+			if (index > -1) {
+				String extension = fileName.substring(index);
+				if (".js".equalsIgnoreCase(extension.toLowerCase())) {
```
mickaelistria, line 63: We need to use ContentType rather than filenames. Content-Types have hierarchy, can be extended, customized by other plugins and users...
So we should just get the Wild Web Developer content type with `PlatformUI.getContentTypeManager().getContentType("org.eclipse.wildwebdeveloper.js")` then ideally compare it with the content-type given as method parameters, or -if it doesn't work- try `contentType.isAssociatedWith(fileName)`.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
```diff
@@ -0,0 +1,73 @@
...
+	public IEditorDescriptor[] overrideEditors(IEditorInput editorInput, IContentType contentType,
+			IEditorDescriptor[] editorDescriptors) {
+		return editorInput != null ? overrideEditors(editorInput.getName(), contentType, editorDescriptors)
+				: editorDescriptors;
+	}
+
+	@Override
+	public IEditorDescriptor[] overrideEditors(String fileName, IContentType contentType,
+			IEditorDescriptor[] editorDescriptors) {
+		return Arrays.stream(editorDescriptors).map(descriptor -> getEditorDescriptorForFile(descriptor, fileName)).toArray(size -> new IEditorDescriptor[size]);
```
mickaelistria, line 40: you can replace `size -> new IEditorDescriptor[size])` by `IEditorDescriptor[]::new`

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
vrubezhny: I'll try to use this approach with `contentType.isAssociatedWith(fileName)`.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
vrubezhny: This code is is dropped in favour of overriding the `overrideDefaultEditor(...)` methods

vrubezhny:
The PR is changed to `override `overrideDefaultEditor(...)` instead of `overrideEditors(..)` as it doesn't change the list of available editors for JS files, but makes the Generic text Editor always to be a default one for these files.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
```diff
@@ -0,0 +1,66 @@
...
+		if (!ExtensionBasedTextEditor.GENERIC_EDITOR_ID.equals(defaultDescriptor.getId())
+				&& fileName != null && !fileName.isEmpty()) {
+			int index = fileName.lastIndexOf('.');
+			if (index > -1) {
+				String extension = fileName.substring(index);
+				if (".js".equalsIgnoreCase(extension.toLowerCase())) {
+					IEditorRegistry editorReg = PlatformUI.getWorkbench()
+							.getEditorRegistry();
+					
+					IEditorDescriptor genericEditorDescriptor = editorReg.findEditor(ExtensionBasedTextEditor.GENERIC_EDITOR_ID);
```
mickaelistria, line 54: This could be made a final field, defined int he constructor, as we're actually reusing it all the time.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/JSEditorAssociationOverride.java
```diff
@@ -0,0 +1,66 @@
...
+					if (genericEditorDescriptor != null) {
+						if (!defaultDescriptorMap.containsKey(fileName)) {
+							defaultDescriptorMap.put(fileName, new GenericEditorWithContentTypeIcon(fileName, genericEditorDescriptor));
+						}
+						return defaultDescriptorMap.get(fileName);
```
mickaelistria, line 59: I think you can skip the map an just return `new GenericEditorWithContentTypeIcon`.

vrubezhny: Then it will be a memory leak. We shouldn't create yet another descriptor for an editor if we already have one for a file.

mickaelistria: The map is the memory leak as there is nothing that ever cleans it.
But you're right, we can avoid instantiating too many `GenericEditorWithContentTypeIcon`. Actually, as the fileName is simply used to resolve the content-type and we already know the interesting content-type here, we can even just instantiate it once and store it as a final field.
```java
public class JSEditorAssociationOverride  .... {
  private final IEditorDescriptor genericEditorWithJSIcon;

  public JSEditorAssociationOverride() {
     IEditorDescriptor genericEditor = PlatformUI.getEditorRegistry().findEditor(ExtensionBasedTextEditor.GENERIC_EDITOR_ID);
     this.genericEditorWithJSIcon = new GenericEditorWithContentTypeIcon("blah.js", genericEditor); // name doesn't matter as long as it match Wild Web Developer JS content-type for icon resolution
```

and then it becomes simpler to implement `overrideDefaultEditor` as a `return wildWebDeveloperJSContentType.equals(contentType) || wildWebDeveloperJSContentType.isAssociatedWith(fileName) ? genericEditorWithJSIcon : super.overrideDefaultEditor(...));`

vrubezhny: yes, if we're not going to have different icons for JS files we can get rid of map and instantiate the `wildWebDeveloperJSContentType` and  `genericEditorWithJSIcon` in the constructor.

mickaelistria:
Thanks Victor!

AObuchow:
Thank you Victor @vrubezhny!

level420:
@mickaelistria I'm in need of using another javascript editor, not the wwd one. Currently with release wwd 0.10.0.202006021616 when clicking on a javascript file, the "generic editor" always gets precedence, even if I've configured the other javascript editor as default under Preferences -> Editors -> File Associations -> *.js where the other editor is set as default.
Of course I can use "Open with" for every *.js file, but that's annoying.
And to be honest: I don't understand if this PR caused this behavior or if it is the fix.

mickaelistria:
I think the PR here may be a cause. Can you please open a separate bug about this issue?

level420:
@mickaelistria done with #448

PR #417, opened 2020-05-05
Remove auto close for < character in XML language configuration

angelozerr:
This PR avoids generating > when < is typing which is really annoying when your are typing XML.


mickaelistria:
I don't find this auto-close particularly annoying. Can you elaborate about specfic use-cases that are made much worse thanks to it.
One metric we can look at is the number of keystrokes to write some XML. The auto-insert here seems to be on par with no insert at all, as user usually has to type the right-arrow to move cursor avec the `>` instead of typing `>` directly. If there are cases where auto-close consumes more keystrokes and no obvious case where it consumes less, we can get rid of it.

mickaelistria:
Let's close by lack of consensus

Issue #418, opened 2020-05-05
WWD cannot find node at /usr/local/bin/node on macOS


jpmat296:
Since I've installed node@12 thanks to Homebrew, Node is not installed in standard location. In such case, I think symbolic link in /usr/local/bin is supposed to work. It doesn't for me since I get popup "Node.js version could not be determined".

Here is the symlink:
```
⇒  ls -l /usr/local/bin/node
lrwxr-xr-x  1 jpmat  admin  31  5 May 10:23 /usr/local/bin/node -> /usr/local/opt/node@12/bin/node
```

Here are versions I use:
- Eclipse IDE 2020-03
- WWD snapshot 0.9.1.202005041657 

I applied this workaround:
```
cd /Applications/Eclipse.app/Contents/Eclipse
echo -Dorg.eclipse.wildwebdeveloper.nodeJSLocation=$(which node) >> eclipse.ini
```

mickaelistria:
Does it work with the workaround?
Is node part of the accessible PATH for the Eclipse IDE? IIRC, Mac is specifically annoying on that topic of sharing environment with apps, so maybe node isn't visible in the PATH.
Or is the issue more generally that node as symlink doesn't work (even when visible in PATH)?
Anyway, if there is no better way to handle that, we would welcome a contribution that adds some heuristics of more default locations to look at for node in case it wasn't found otherwise.

jpmat296:
Yes, the workaround works with symlink or not.

Actually PATH cannot be used on MacOS. When Eclipse is started by the icon, PATH is not well defined.

But if I start Eclipse by terminal, then it works:
```
⇒ which node
/usr/local/opt/node@12/bin/node
⇒  cd /Applications/Eclipse.app
⇒  ./Contents/MacOS/eclipse
# Works fine
```

What is not clear to me is why WWD default location detection doesn't work:
https://github.com/eclipse/wildwebdeveloper/blob/4030c2dc9c68dd9fe97c3429fd519153cdb273cc/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/InitializeLaunchConfigurations.java#L95-L98

I will try to debug.

jpmat296:
I've submitted PR.

Issue was not in getDefaultNodePath() but in which(node) code. Please see PR #422 for more details.

mickaelistria:
[...]

PR #422, opened 2020-05-06
Issue #418: Use default user shell on macOS to find node

jpmat296:
This PR change which(node) behavior only for macOS.

On macOS, Bash is not the default shell since macOS 10.15. Now it's /bin/zsh.

Before WWD which(node) implementation used always /bin/bash to find NodeJS path which fails when Bash is not configured.

Now the new code first find the shell for current user and then find NodeJS path. It should always succeed to find NodeJS, as long as user has configured NodeJS in its shell.

### DIFF
--- a/org.eclipse.wildwebdeveloper.feature/feature.xml
+++ b/org.eclipse.wildwebdeveloper.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.9.1.qualifier"
+      version="0.10.1.qualifier"
       provider-name="Eclipse Wild Web Developer project"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/org.eclipse.wildwebdeveloper.feature/pom.xml
+++ b/org.eclipse.wildwebdeveloper.feature/pom.xml
@@ -7,7 +7,7 @@
 		<version>0.5.0-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-feature</packaging>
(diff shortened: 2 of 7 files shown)


mickaelistria:
I'm curious about compatilibity with old mac. I see 10.15 is the latest one but previous releases are still supported. Is your patch supposed to work with older releases?

jpmat296:
Before 10.15, Bash was the default Shell. My code detects that thanks to "dscl" command to find the current user shell.

And "dscl" is available since a long time on macOS (see [here](https://www.techrepublic.com/article/pro-tip-5-more-commands-mac-admins-should-know/) for example).

I've not tested myself. I can test if you think it's better.

mickaelistria:
If you can test without too much trouble, I'd be grateful if you can do it.
Also, can you please bump the bundle version (in MANIFEST and pom.xml) to 0.5.7 ? I think you'll also need to bump the feature version by +0.0.1.

mickaelistria:
I recently push a change and will handle the rebase on merge. So don't worry to much about the versions no more, I'll take care of it during rebase.
I'll merge as soon as you confirm it's fine on older Mac versions (if you're still able to test that).

jpmat296:
Ok I have done the test successfully with :
- macOS 10.14.6
- macOS 10.13.6
- macOS 10.12.6

FYI during debugging, I have detected which(node) is executed often. It should be called only once.

mickaelistria:
Thanks, I'm going to rebase and merge.
Please open another issue (or even a PR) about your other findings.

mickaelistria:
This was rebased and merged as 7b793ca58281e58db404890512a42d746cf35c68 . Thanks a lot!

Issue #429, opened 2020-05-17
[feature request] JSON schema association via extension point


zulus:
I would like to replace WTP JSON by WildWebDeveloper in Eclipse PDT Composer editor (eclipse/pdt#54). Only feature I miss is no ability to associate composer.json with correct schema.

In `JSonLanguageServer.java` I found only TODO tag. I'll be happy to have something similar to `org.eclipse.wst.json.core.schemaCatalogContributions`.

Example from WTP:

```XML
<extension
         point="org.eclipse.wst.json.core.schemaCatalogContributions">
         <schemaCatalogContribution id="http://schemastore.org/">
            <schema name="composer.json"
                  description="%schema.description"
                  fileMatch="composer.json"
                  url="http://json.schemastore.org/composer"
                  uri="schemastore/composer" />
         </schemaCatalogContribution>
</extension>
```

mickaelistria:
See also https://github.com/eclipse/wildwebdeveloper/pull/316

mickaelistria:
Fixed with https://github.com/eclipse/wildwebdeveloper/pull/430

PR #430, opened 2020-05-17
Register JSON Schema via extension point

zulus:
Simple implementation of eclipse/wildwebdeveloper#429.

Example usage:

```xml
<extension point="org.eclipse.wildwebdeveloper.json.schema">
      <schema pattern="composer.json" url="http://json.schemastore.org/composer" />
      <schema pattern="/tsconfig.*.json" url="http://json.schemastore.org/tsconfig" />
</extension>
```

Note: patch also contain macOS test fix

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/pom.xml
+++ b/org.eclipse.wildwebdeveloper.tests/pom.xml
@@ -31,6 +31,7 @@
 				<artifactId>tycho-surefire-plugin</artifactId>
 				<version>${tycho-version}</version>
 				<configuration>
+					<argLine>${tycho.testArgLine} ${ui.test.vmargs}</argLine>
 					<useUIThread>true</useUIThread>
 					<useUIHarness>true</useUIHarness>
 				</configuration>
--- a/org.eclipse.wildwebdeveloper/schema/json.schema.exsd
+++ b/org.eclipse.wildwebdeveloper/schema/json.schema.exsd
@@ -18,7 +18,7 @@
       </annotation>
       <complexType>
          <sequence minOccurs="1" maxOccurs="unbounded">
-            <element ref="file"/>
+            <element ref="schema"/>
          </sequence>
          <attribute name="point" type="string" use="required">
             <annotation>
@@ -47,21 +47,6 @@
       </complexType>
    </element>
 
-   <element name="file">
-      <complexType>
-         <sequence minOccurs="1" maxOccurs="unbounded">
-            <element ref="schema"/>
-         </sequence>
-         <attribute name="name" type="string" use="required">
-            <annotation>
-               <documentation>
-                  File name pattern
-               </documentation>
-            </annotation>
-         </attribute>
-      </complexType>
-   </element>
-
    <element name="schema">
       <annotation>
          <documentation>
@@ -69,6 +54,13 @@
          </documentation>
       </annotation>
       <complexType>
+         <attribute name="pattern" type="string" use="required">
+            <annotation>
+               <documentation>
+                  
+               </documentation>
+            </annotation>
+         </attribute>
       </complexType>
    </element>
 
@@ -77,7 +69,7 @@
          <meta.section type="since"/>
       </appinfo>
       <documentation>
-         [Enter the first release in which this extension point appears.]
+         0.5.8
       </documentation>
    </annotation>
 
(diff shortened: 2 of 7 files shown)


org.eclipse.wildwebdeveloper/schema/json.schema.exsd
```diff
@@ -0,0 +1,112 @@
...
+      <complexType>
+      </complexType>
+   </element>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="since"/>
+      </appinfo>
+      <documentation>
+         [Enter the first release in which this extension point appears.]
```
mickaelistria, line 80: Please add content

org.eclipse.wildwebdeveloper/schema/json.schema.exsd
```diff
@@ -0,0 +1,112 @@
...
+      </documentation>
+   </annotation>
+
+   <annotation>
+      <appinfo>
+         <meta.section type="examples"/>
+      </appinfo>
+      <documentation>
+         [Enter extension point usage example here.]
```
mickaelistria, line 89: Please add content

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/json/JSonLanguageServer.java
```diff
@@ -108,6 +115,21 @@ private void fillSchemaAssociationsForTypeScript(Map<String, List<String>> assoc
 	private void fillSchemaAssociationsForOmnisharp(Map<String, List<String>> associations) {
 		associations.put("/omnisharp.json", Arrays.asList("http://json.schemastore.org/omnisharp"));
 	}
+	
+	private void fillSchemaAssociationsFromExtensionPoint(Map<String, List<String>> associations) {
+		IConfigurationElement[] conf = Platform.getExtensionRegistry().getConfigurationElementsFor(SCHEMA_EXT);
+		for (IConfigurationElement el : conf) {
+			associations.put(el.getAttribute(NAME_ATTR), collectSchemaURL(el.getChildren(SCHEMA_ELEMENT)));
```
mickaelistria, line 122: why not schema as attribute?
Also, what it 2 distinct plugin want to add different schema for the same name/pattern? In this case, I would expect that the list would be joined.

zulus: Currently it will be replaced. As I saw in hardcoded part, it's always registered as list so I prepared extension point in same way.

I can reimplement it as append/merge. I'm also thinking about rename `name` to `pattern`.

mickaelistria: Renaming the name to pattern seems good.
Allowing to merge extensions for same pattern seems good too.
As for children elements vs attributes, I don't have a strong preference. I'll let you defend the structure you prefer.

zulus: I updated documentation and simplify extension schema

org.eclipse.wildwebdeveloper/schema/json.schema.exsd
```diff
@@ -0,0 +1,107 @@
...
+               </appinfo>
+            </annotation>
+         </attribute>
+      </complexType>
+   </element>
+
+   <element name="schema">
+      <annotation>
+         <documentation>
+            URL as value
```
mickaelistria, line 53: It seems to me the PDE extension editor isn't really smart enough dealing with direct text values (ie the form editor doesn't suggest entries for those). Unless I'm mistaken, it's probably worth sticking to an attribute.

zulus: Yeah, it's also not possible to set restrictions on element value. In theory I could set pattern and force valid URL in attribute.

org.eclipse.wildwebdeveloper.tests/pom.xml
```diff
@@ -31,6 +31,7 @@
 				<artifactId>tycho-surefire-plugin</artifactId>
 				<version>${tycho-version}</version>
 				<configuration>
+					<argLine>${tycho.testArgLine} ${ui.test.vmargs}</argLine>
```
mickaelistria, line 34: This is making build fail on Linux, see automated CI review.
`ui.test.vmargs` must be set as empty by default, and overriden in some mac-specific profile

zulus: Done

mickaelistria:
While CI is processing the patch, do you think you can add a note aobut it in the RELEASE_NOTES.md as part of this commit?

mickaelistria:
The patch is ready to be merged (test failure is not related and happens sometimes on slower CI slaves). Do you prefer I merge it now and you add notes to RELEASE_NOTES later, or do you prefer I wait for an amended version that includes the RELEASE_NOTES in the same commit?

zulus:
I'll update release_notes. I can even squash all commits into one if you want.

mickaelistria:
Squashed and merged as 73f86c23e3401e4545a1d22899c668c8515693c6 
Thanks!

PR #435, opened 2020-06-02
Draft: [Waiting for CQ] Update language servers

vrubezhny:
The following LSs will be updated to the latest versions:
typescript@3.8.3
typescript-plugin-css-modules@2.3.0
yaml-language-server@0.8.0
firefox-debugadapter@2.8.0
typescript-eslint-language-service@3.0.0
eslint@7.1.0
@typescript-eslint/eslint-plugin@3.1.0
@typescript-eslint/parser@3.1.0
@angular/language-service@9.1.9
@angular/language-server@0.901.9

Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

PR #437, opened 2020-06-04
Account pom.xml change in qualifier computation

mickaelistria:
As most embedded deps are configured in pom.xml and the pom files don't
change much otherwise, let's take the pom.xml into account for qualifier
computation so important changes lead to new version (preventing from
further baseline replacement).
Also remove package-lock.json on clean.

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for WildWebDeveloper
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper.tests
-Bundle-Version: 0.4.4.qualifier
+Bundle-Version: 0.4.5.qualifier
 Bundle-Vendor: Eclipse Wild Web Developer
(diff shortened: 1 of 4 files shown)


jabby:
@mickaelistria I'm not sure this is a good idea to remove package-lock.json. Maybe it's ok. IIUC you want to use only version number of npm dependencies in the pom.xml. As I can see, we use exact version for those dependencies and it is good (at least, it's my opinion). But what happen for transitive dependencies?

In JS world, most of developers don't fix the version number of their dependencies but use range dependencies. So between two builds if there is no lock file and one of transitive dependency is updated, we can get another version of this transitive dependency. Is it compliant with our CQs? 

Maybe it will be ok and I suggest to check if there is no problem between two builds when some transitive dependencies will have new versions.

jabby:
I miss this comment here https://github.com/eclipse/wildwebdeveloper/pull/435#issuecomment-638287858
It makes sense.

mickaelistria:
What we're trying to achieve is that we get the bundle updated when we update dependencies explicitly. Because there is no Git change when transitive deps change, the bundle is replaced by previous build (with same fully qualified version). So we're good here: there is no chance that we publish updated transitive deps unless we explictely change something.

jabby:
Good to know. I am not really familiar with the baseline validation of Eclipse. I know it's a shame. The comment I linked is quite explicit and I am now quite confident with this change. 
Thanks for your explanation.

Issue #144, opened 2019-03-14
Better support TypeScript debugging


mickaelistria:
TypsScript debugging needs to be a better citizen:

1. it was never tried nor reported as working
2. how to should be documented
3. Tests added
4. Overall workflow improved

mickaelistria:
The node debugger seems to have some capabilities for such files, by relying on sourceMaps. However, sourceMaps support is currently disabled (to prevent from other failures in the debug adapter). Work should be done in trying to re-enable sourceMaps in the LS so we could try to run directly .ts files.

AObuchow:
For the case of debuging a website written in TypeScript, see https://github.com/eclipse/wildwebdeveloper/blob/master/TIPS_and_FAQ.md#debug-a-website-developed-in-typescript

vrubezhny:
An issue is reported on lsp4e: Bug 564070 - Launch goes into an endless loop when a launch is failed: https://bugs.eclipse.org/bugs/show_bug.cgi?id=564070 and a PR is provided:  https://git.eclipse.org/r/164428

PR #443, opened 2020-06-11
Better support TypeScript debugging #144

vrubezhny:
No TSConfig file case s fixed

Fixes: #144 


mickaelistria:
The button label should be either "Edit tsconfig.json" or "Create tsconfig.json" depending on the case.
Also, the multiple folder/file creations should be wrapped into a single WorkspaceOperation to avoid sending more notifications than necessary.

vrubezhny:
> The button label should be either "Edit tsconfig.json" or "Create tsconfig.json" depending on the case.
> Also, the multiple folder/file creations should be wrapped into a single WorkspaceOperation to avoid sending more notifications than necessary.

Fixed

mickaelistria:
Thanks!

Issue #23, opened 2017-08-11
Embed Node.js Runtime


angelozerr:
Provide embed node.js runtime like I have done for typescript.java. It will give the capability to consumes LSP without installing node.js (VSCode does like this I think).

If you are interested, I could do a PR.

mickaelistria:
How likely it is that a frontend developer doesn't have nodejs installed?
I don't think it's worth the effort.


-- 
Mickael Istria
Eclipse IDE <https://www.eclipse.org/downloads/eclipse-packages/>
developer, at Red Hat Developers <https://developers.redhat.com/> community

angelozerr:
If you are developping JEE Application and you wish to benefit with HTML, CSS features you don't need node.js. IMHO, I think Bluesky should work without any other program installation like node.js. VSCode works like this I think.

mickaelistria:
There are package managers such as rpm for that. I don't think the burden should be carried by BlueSky, especially as it would drastically grow the download size and may lead users to inconsistencies ("why don't Eclipse and command-line behave the same...?  [3 hours later] Ok, it's not using the same node.js!").
A simple warning pop-up if node.js is not available or version is too old should be enough.

angelozerr:
> There are package managers such as rpm for that. 

Take a sample. I have a computer where node.js is not installed and I want to develop HTML, CSS, JavaScript application:

 * I install VSCode, it works.
 * I install Eclipse IDE + BlueSky, it doesn't work. I don't need node.js, so I telll me why I must install node.js. I give up BlueSky and uses VSCode. 

None offense with your idea, I would like just to give you some feedback with typescript.java where user like the embed node.js feature.

> it would drastically grow the download size

The update site provides (like typescript.java) an optional installation of the embed node.js

mickaelistria:
Ok, so it could simply be a change in https://github.com/mickaelistria/eclipse-bluesky/blob/master/org.eclipse.bluesky/src/org/eclipse/bluesky/InitializeLaunchConfigurations.java#L56 to use alternative strategies:
* If node is in PATH, use it (current)
* If not and node is found in the Eclipse IDE installation (we could/should consume the node.js runtime packaged into some Eclipse plugin.

Doesn't JSDT already provided an embedded node.js that we could reuse (instead of writing yet-another-bundle including nodejs?) - cc @ibuziuk
Or maybe we could simply reuse the bundle that's in typescript.java

ibuziuk:
[...]

PR #458, opened 2020-06-30
Embed Node.js Runtime #23

vrubezhny:
Fixes: #23


org.eclipse.wildwebdeveloper.embedder.node/src/org/eclipse/wildwebdeveloper/embedder/node/InitializeLaunchConfigurations.java
```diff
@@ -14,21 +14,26 @@
  *   Pierre-Yves B. - Issue #238 Why does wildweb do "/bin/bash -c which node" ?
  *   Pierre-Yves B. - Issue #268 Incorrect default Node.js location for macOS
  *******************************************************************************/
-package org.eclipse.wildwebdeveloper;
+package org.eclipse.wildwebdeveloper.embedder.node;
```
mickaelistria, line 17: should we rename the class by the way, to something like `NodeJSManager` or `NodeJSAccessor` ?

vrubezhny: I like `NodeJSManager` or `NodeJSRuntimeProvider`. But the first name is better as it's shorter.

org.eclipse.wildwebdeveloper.embedder.node/src/org/eclipse/wildwebdeveloper/embedder/node/InitializeLaunchConfigurations.java
```diff
@@ -70,13 +102,12 @@ public static String getNodeJsLocation() {
 		return null;
 	}
 
-	public static String which(String program) {
-
+	public static File which(String program) {
```
vrubezhny, line 105: Should we make `which()` to search for the specified program in the embedded NodeJS executable installation folder first, then using other possible paths?

vrubezhny:
Last changes:
- Linux and MacOS fragments are switched to use *.tar.gz formats of NodeJS distribution as it gives x2.6 times faster decompression even given bigger size of the distribution archives. 
- InitializeLaunchConfigurations renamed to NodeJSManager;
- which() is made to search first inside embedded NodeJS installation folder;
- Node Debug is configured to use embedded NodeJS (if embedding takes place) instead of any occasionally installed NodeJS distribution.

mickaelistria:
All that seems really good!
I'm merging it so we can more easily test. We need to test it on environment where node.js is absent. I'll probably tweak the Jenkinsfile later to have tests running without node.js in path.

Issue #461, opened 2020-07-02
Cannot use provided schemas


apupier:
when trying to use a specific schemas provided in preferences, when openign the file and asking for completion, there is this error:
```
eclipse.buildId=4.15.0.I20200305-0155
java.version=1.8.0_252
java.vendor=Oracle Corporation
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Framework arguments:  -product org.eclipse.epp.package.committers.product
Command-line arguments:  -data /home/apupier/ws/camel-ls -os linux -ws gtk -arch x86_64 -product org.eclipse.epp.package.committers.product

This is a continuation of log file /home/apupier/ws/camel-ls/.metadata/.bak_5.log
Created Time: 2020-07-02 16:50:27.000

org.eclipse.lsp4e
Error
Thu Jul 02 16:56:59 CEST 2020
org.eclipse.lsp4j.jsonrpc.ResponseErrorException: Request textDocument/completion failed with message: entry.getSchemas(...) is not a function or its return value is not iterable

java.util.concurrent.ExecutionException: org.eclipse.lsp4j.jsonrpc.ResponseErrorException: Request textDocument/completion failed with message: entry.getSchemas(...) is not a function or its return value is not iterable
	at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:357)
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1908)
	at org.eclipse.lsp4e.operations.completion.LSContentAssistProcessor.computeCompletionProposals(LSContentAssistProcessor.java:105)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$10(AsyncCompletionProposalPopup.java:351)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.text.contentassist.AsyncCompletionProposalPopup.lambda$9(AsyncCompletionProposalPopup.java:350)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1604)
	at java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1596)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)
[...]

PR #488, opened 2020-07-21
Cannot use provided schemas #461

vrubezhny:
Fixes: #461

The PR provides dynamic schemas and WorkspaceFolder list configuration (by listening for the Preferences/Workspace changes respectively).

### DIFF
--- a/org.eclipse.wildwebdeveloper/pom.xml
+++ b/org.eclipse.wildwebdeveloper/pom.xml
@@ -49,7 +49,7 @@
 							<goal>wget</goal>
 						</goals>
 						<configuration>
-							<url>https://redhat.gallery.vsassets.io/_apis/public/gallery/publishers/msjsdiag/vsextensions/debugger-for-chrome/4.12.8/vspackage</url>
+							<url>https://msjsdiag.gallery.vsassets.io/_apis/public/gallery/publisher/msjsdiag/extension/debugger-for-chrome/4.12.8/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage</url>
 							<outputFileName>chromeDebugAdapter.zip</outputFileName>
 							<unpack>true</unpack>
 							<outputDirectory>${project.build.directory}/chrome-debug-adapter</outputDirectory>
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
@@ -21,9 +21,16 @@ import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.jface.util.IPropertyChangeListener;
+import org.eclipse.jface.util.PropertyChangeEvent;
+import org.eclipse.lsp4e.LanguageServersRegistry;
+import org.eclipse.lsp4e.LanguageServersRegistry.LanguageServerDefinition;
+import org.eclipse.lsp4e.LanguageServiceAccessor;
 import org.eclipse.lsp4e.server.ProcessStreamConnectionProvider;
 import org.eclipse.lsp4j.DidChangeConfigurationParams;
+import org.eclipse.lsp4j.InitializeResult;
 import org.eclipse.lsp4j.jsonrpc.messages.Message;
+import org.eclipse.lsp4j.jsonrpc.messages.ResponseMessage;
 import org.eclipse.lsp4j.services.LanguageServer;
 import org.eclipse.wildwebdeveloper.Activator;
 import org.eclipse.wildwebdeveloper.embedder.node.NodeJSManager;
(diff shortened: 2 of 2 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
```diff
@@ -13,27 +13,108 @@
...
+				Map<String, Object> options = new HashMap<>();
+				options.put(SETTINGS_KEY, settings);
+				DidChangeConfigurationParams params = new DidChangeConfigurationParams(settings);
+				LanguageServiceAccessor.getActiveLanguageServers(null).stream().filter(server -> yamlLsDefinition.equals(LanguageServiceAccessor.resolveServerDefinition(server).get()))
+					.forEach(ls -> ls.getWorkspaceService().didChangeConfiguration(params));
+			}
+		}
+ 	};
+ 	
+ 	private static final IResourceChangeListener wsFoldersListener = new IResourceChangeListener() {
```
mickaelistria, line 79: This is more something to implement in LSP4E directly. But I believe before that, it would be more interesting to investigate whether the Yaml-LS couldn't just ignore the rootURI at that time.

vrubezhny: I didn't found this in LSP4E and I don't know if this is something to be implemented there. I see this like something that is YAML LS specific as the Workspace Folders sent to the LS are only Project folders (without their internal sub-folders) - which makes less overload in communication between client and server. 
And I believe, YAML LS can do the same by itself, if it was properly configured at start by LSP4E (see: https://github.com/eclipse/wildwebdeveloper/issues/461#issuecomment-661559744 on `rootUri`)

On why YAML LS "couldn't just ignore the riootUri" - if it will ignore that - where and how it can find the workspace root then? And it is probably has such a problem because LSP4E configures `rootUri` to point to one of the Workspace projects and not to the Workspace root (It's probably OK for VSCode, but not for Eclipse's Workspace, imho). That's why these Workspace Folders (projects) now are required by LS.

mickaelistria:
What I think is wrong-ish somewhere in yaml-ls is that it require workspaceFolders or other things for schemas while similar LS like the JSON or the XML ones are totally independent of the workspace structure and don't cause such issues. I think the Yaml-LS should mimic those other ones.
As for adding automatic support for workspace folders to LSP4E upon server initialization, and update upon addition/removal of projects, that would be welcome in LSP4E.
I don't think any issue actually belongs (nor should be fixed) in Wild Web Developer here, it's more partly LS/partly LSP4E.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
```diff
@@ -53,27 +87,40 @@ public YAMLLanguageServer() {
...
 	}
-	
+
 	@Override
 	public String toString() {
 		return "YAML Language Server: " + super.toString();
 	}
+
+	@Override
+	protected void finalize() throws Throwable {
```
mickaelistria, line 122: finalize is deprecated and shouldn't be used at all. There is a `stop` method you can override instead.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
```diff
@@ -21,19 +21,53 @@
...
+	private static final String YAML_KEY = "yaml";
+	private static final String VALIDATE_KEY = "validate";
+	private static final String COMPLETION_KEY = "completion";
+	private static final String HOVER_KEY = "hover";
+	private static final String SCHEMAS_KEY = "schemas";
 	
-	private String cachedSchema;
+ 	private static final IPreferenceStore store = Activator.getDefault().getPreferenceStore();
+ 	private static final LanguageServerDefinition yamlLsDefinition = LanguageServersRegistry.getInstance().getDefinition("org.eclipse.wildwebdeveloper.yaml");
+ 	private static final IPropertyChangeListener psListener = new IPropertyChangeListener() {
```
mickaelistria, line 52: Why would we need to change those preferences to false? I suggest we simply get rid of the preference listener, and hardcode the values to yes, until someone has a good reason to disable it.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
vrubezhny: Ah. Sure. Thanks for the point

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
vrubezhny: That's not about enabling or disabling schemas processing. It's about the loading schema itself. Listener updates the schema on server when a user changed it YAML Schemas preference page. So if a schemas preference is not empty (user defined or changed some schemas) we're sending the new schemas to the server and enabling it to work with schemas.
So, we need this listener.
The only point missed here, probably, is that YAML Preference page has no schemas defined, it should disable schemas processing on the server. (No schemas loaded - no processing is needed)

mickaelistria: OK, got it.

org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -49,7 +49,7 @@
 							<goal>wget</goal>
 						</goals>
 						<configuration>
-							<url>https://redhat.gallery.vsassets.io/_apis/public/gallery/publishers/msjsdiag/vsextensions/debugger-for-chrome/4.12.8/vspackage</url>
+							<url>http://127.0.0.1:8080/msjsdiag.debugger-for-chrome-4.12.8.vsix</url>
```
mickaelistria, line 52: Wrong URI.

vrubezhny: Yeah. sorry. fixed

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/yaml/YAMLLanguageServer.java
vrubezhny: I was wrong on these preferences. Independently on existence of schemsa  these preferences are to be set tu true as yaml file still needs to be validated. This was the reason of failing `TestLanguageServers.testYAMLFile()` test case.

PR #498, opened 2020-08-26
Build with Tycho 2.0.

akurtakov:
Change-Id: I1bdf196548ec3ffae27430718bbc0444e96b2152

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,59 +3,24 @@ pipeline {
 		timeout(time: 30, unit: 'MINUTES')
 		buildDiscarder(logRotator(numToKeepStr:'10'))
 	}
-  agent {
-    kubernetes {
-      label 'wildwebdeveloper-buildtest-pod-f31-take6'
-      defaultContainer 'jnlp'
-      yaml """
-apiVersion: v1
-kind: Pod
-spec:
-  containers:
-  - name: container
-    image: docker.io/mickaelistria/fedora-gtk3-mutter-java-node:31-gtk3.24
-    imagePullPolicy: Always
-    tty: true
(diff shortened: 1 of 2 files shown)


gautier-desaintmartinlacaze-cbp:
@mickaelistria I think we need a new version of docker container with Java 11 instead of Java 8. Can you create a new version of docker container?

jabby:
@mickaelistria I think we should update our build container. Actually we use Java 8 for building but with the new Eclipse IDE requirement we should build with Java 11.
As you are the owner of the docker container on docker hub, can you upload a new version?

mickaelistria:
Actually, now that we have some node embedder, we could switch back to default images/agents like `centos-latest` provided by the Eclipse infra and use the built-in tools. As we don't really need node any more, it should be good to test.
However, it would then mean we'd test against the embedded node (good) and that we wouldn't test any more against the system one (bad). I don't have a clear preference for one or the other, so if using embedded node simplifies CI, then it's IMO the better choice.

mickaelistria:
FWIW, there is no automated stream from dockerfile (in Git repo) to Jenkinsfile, so to test the new docker image, it requires to build and publish it, and update the Jenkinsfile to use the new image. If we only test embedded node.js, then we can just drop the docker image and use
```
	agent {
		label "centos-latest"
	}
	tools {
		maven 'apache-maven-latest'
		jdk 'openjdk-jdk11-latest'
	}
```

akurtakov:
I have added additional patch that makes sure java 11 is installed in the image and used in the jenkinsfile before reading your comment. Using default image might be good idea too.

akurtakov:
I'm closing this one as we will still need our image so we can npm install things at build time and it will be multi step effort.

PR #526, opened 2020-09-18
Switch to yaml-language-server 0.11.0

### LINKED ISSUE #244: [Firefox Debug] Pre-set file in Launch Configuration when editor has focus
1. Open some HTML file in an Eclipse IDE editor, ensure editor is active (has focus)
2. Debug As > Launch Firefox Debugger > New Configuration
EXPECTED: the current file path is pre-set in the Launch Configuration
GOT: Empty file path

### LINKED ISSUE #372: YAML should not support tabs.
According to this: https://yaml.org/spec/1.2/spec.html#id2777534
`In YAML block styles, structure is determined by indentation. In general, indentation is defined as a zero or more space characters at the start of a line.

To maintain portability, tab characters must not be used in indentation, since different systems treat tabs differently. Note that most modern editors may be configured so that pressing the tab key results in the insertion of an appropriate number of spaces.

The amount of indentation is a presentation detail and must not be used to convey content information.

Each node must be indented further than its parent node. All sibling nodes must use the exact same indentation level. However the content of each sibling node may be further indented independently..`
Wild web developer allows us to use YAML files with tabs.
https://imgur.com/58hVbWM
Is there any change to do in generic editor to allows convering tabs to space?

jeffmaury:
Fixes #372 

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

### DIFF
--- a/org.eclipse.wildwebdeveloper/pom.xml
+++ b/org.eclipse.wildwebdeveloper/pom.xml
@@ -121,7 +121,7 @@
 								<arg>${project.build.directory}/vscode-css-languageserver-1.0.0.tgz</arg>
 								<arg>${project.build.directory}/vscode-html-languageserver-1.0.0.tgz</arg>
 								<arg>${project.build.directory}/vscode-json-languageserver-1.2.3.tgz</arg>
-								<arg>yaml-language-server@0.10.1</arg>
+								<arg>yaml-language-server@0.11.0</arg>
 								<arg>firefox-debugadapter@2.9.0</arg>
 								<arg>${project.build.directory}/debugger-for-chrome-4.12.8.tgz</arg>
 								<arg>typescript-eslint-language-service@3.0.0</arg>


jeffmaury:
Tested and it works with a schema and code completion.
However, if I enter the following in an empty file:

`mytag:`

and then ENTER then the cursor is on next line and column 0 (no indentation at all) but that was perhaps the same before;. Maybe @evidolob can comment

jabby:
@jeffmaury Are you sure it's about #244 ?

jeffmaury:
> 
You are right, wrong repo issue link, I updated it

evidolob:
> However, if I enter the following in an empty file:
> 
> `mytag:`
> 
> and then ENTER then the cursor is on next line and column 0 (no indentation at all) but that was perhaps the same before;.

Yaml-ls does't do indentation on text edits, we fix indentation on codecompletion.

mickaelistria:
@jeffmaury Can you please rebase it?

mickaelistria:
Opened CQ https://dev.eclipse.org/ipzilla/show_bug.cgi?id=22605

mickaelistria:
Merged as 50b320cd65e5ebcb5e57ad1c8d39458033610db5

PR #531, opened 2020-09-21
Fix HTML formatting enablement 

### LINKED ISSUE #528: HTML Formatting Greyed Out
Hi,

I hope this is just user error but is there a reason why the HTML formatting command is greyed out and not available.

Fresh install of 2020-09 Mac & Windows, same behaviour.

AObuchow:
Fix #528

### DIFF
--- a/org.eclipse.wildwebdeveloper.feature/feature.xml
+++ b/org.eclipse.wildwebdeveloper.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.10.4.qualifier"
+      version="0.10.5.qualifier"
       provider-name="Eclipse Wild Web Developer project"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/org.eclipse.wildwebdeveloper.feature/pom.xml
+++ b/org.eclipse.wildwebdeveloper.feature/pom.xml
@@ -7,7 +7,7 @@
 		<version>0.5.0-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-feature</packaging>
-	<version>0.10.4-SNAPSHOT</version>
+	<version>0.10.5-SNAPSHOT</version>
 	<build>
 		<plugins>
 			<plugin>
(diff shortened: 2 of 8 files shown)


AObuchow:
For some reason I was getting a tycho baseline version should-be-changed error, so I bumpbed the node embdedder version. Going to remove that from the PR in a moment..

AObuchow:
Also, I tested that this works :)

image

mickaelistria:
@AObuchow thanks! Could you please create a test for that (getting the format command with the CommandService and checking its enablement should be enough)?

AObuchow:
> @AObuchow thanks! Could you please create a test for that (getting the format command with the CommandService and checking its enablement should be enough)?

@mickaelistria for sure :) (Although it might take a little bit longer.. I should get back to studying ;) )

Do you happen to know any example tests that show how to retrieve the CommandService? Maybe in LSP4E or Wild Web Developer.

mickaelistria:
`PlatformUI.getWorkbench().getService(ICommandService.class).getCommand("blah")` should work

AObuchow:
> `PlatformUI.getWorkbench().getService(ICommandService.class).getCommand("blah")` should work

OK thank you! Will add a test when I have a moment.

Also the build is failing due to a tycho baseline x.y.z version increment issue. Let me know if I should bump anything to make the build pass.

mickaelistria:
> Also the build is failing due to a tycho baseline x.y.z version increment issue. Let me know if I should bump anything to make the build pass.

I think it's fine if you just rebase your change. Otherwise, whenever you see this message, feel free to bump with +0.0.1 for bug fix, or +0.1.0 for new major feature or new API.

mickaelistria:
While the patch looks good, I'm curious about why the test doesn't fail already? Did you have a chance to look at it?

AObuchow:
> While the patch looks good, I'm curious about why the test doesn't fail already? Did you have a chance to look at it?

I was thinking the same thing, it's odd the existing HTML format test wasn't failing. Unfortunately I didn't have a chance to see which tests were failing before I fixed the bug, so I'm not too sure.

Thanks for the review & merge !

mickaelistria:
It could be that directly calling execute() bypasses the enablement check, and that the LS can accept rename actions even when it doesn't declare the capability.

PR #543, opened 2020-10-08
Support for 'xml/executeClientCommand' protocol extension

BoykoAlex:
Support for handling `xml/executeClientCommand` request.
First attempt to find a matching command registered by one of the Language Servers.
Try to execute eclipse registered command if no LS command found.

A good chunk of client command handling logic (no LS registered command case) taken from LSP4E... Perhaps it'd make sense to make API for this in LSP4E...

Server side PR: https://github.com/eclipse/lemminx/pull/905


org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/CommandExecutor.java
```diff
@@ -0,0 +1,168 @@
+package org.eclipse.wildwebdeveloper.xml.internal;
```
mickaelistria, line 1: Please add EPLv2 copyright header.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/CommandExecutor.java
```diff
@@ -0,0 +1,168 @@
...
+import org.eclipse.ui.handlers.IHandlerService;
+
+@SuppressWarnings("restriction")
+class CommandExecutor {
+	
+	private static final String LSP_COMMAND_CATEGORY_ID = "org.eclipse.lsp4e.commandCategory"; //$NON-NLS-1$
+	private static final String LSP_COMMAND_PARAMETER_TYPE_ID = "org.eclipse.lsp4e.commandParameterType"; //$NON-NLS-1$
+	private static final String LSP_PATH_PARAMETER_TYPE_ID = "org.eclipse.lsp4e.pathParameterType"; //$NON-NLS-1$
+
+	public CompletableFuture<Object> executeClientCommand(String id, Object... params) {
```
mickaelistria, line 47: It looks like all that should be placed in LSP4E, and even that LSP4E does already have such code available. We should instead investigate a better way to expose the LSP4E command handling as API and consume it here.

mickaelistria (changes requested):

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/CommandExecutor.java
kdvolder: Alex is a bit busy so I will continue to push this forwards. I just created a gerrit patch merging in this code with the lsp4e 'CommandExecutor'. 

See: https://git.eclipse.org/r/c/lsp4e/lsp4e/+/170859

I suggest we start by discussion / resolving that patch via gerrit. And once that is done I can then amend this PR here to remove the obsolete wild web copy of CommandExecutor and use the one from lsp4e instead.

kdvolder: Note: you will see that my patch exports internal package. This is probably not what we want in the end. But I wanted to avoid moving the `CommandExecutor` class to another package as that would make it a lot harder to see what was changed in this class. So I think its better we review the change details first, and once we are agreed they are fine, we can then discuss / decide how to expose the api publically.

mickaelistria:
@BoykoAlex @kdvolder While I'm not fundamentally against that change, I believe thas such `executeClientCommand` could be standardized in LSP directly as I imagine plenty of use-cases for it, independently on the language. Is there already a request about it in LSP backlog? I'd like to make sure the idea is tracked at the best place before merging some local workarounds and extensions.

kdvolder:
There's a couple of related LSP issue tickets:

- https://github.com/microsoft/language-server-protocol/issues/430
- https://github.com/microsoft/language-server-protocol/issues/432
- https://github.com/microsoft/language-server-protocol/issues/642

The first one is probably the most closely related, one part of what it needs/asks for is a mechanism for server to request client to execute a command. 

The second one has a discussion around whether commands live in a 'shared/global' name space or whether the commands belong only to a specific client<->server session. 

With a shared namespace one server could define a command and another one could use it (e.g. in code action, or also in a potential `executeClientCommand` request. With 'per session namespace' this wouldn't be possible.

I'm not sure how realistic it is to get this resolved in LSP protocol in the short term, which is why it me be best to move forward with custom protocol extensions such as 'xml/executeClientCommand'. This is not ideal of course.

kdvolder:
BTW: Another option might be to have a 'executeClientCommand' provided by lsp4e. That way at least different eclipse based clients wouldn't all need to implement their own version of it.

mickaelistria:
> There's a couple of related LSP issue tickets

Interesting, however none seems easily actionable, and it looks that they are more brainstorming than actual concrete proposals. Can you please open one about "Introduce a client/executeCommand server->client operation" ? Since you have mostly implemented it for LemMinX already, this proposal might be easily actionable.
While this wouldn't answer every question asked, it would be a great starting point.

> I'm not sure how realistic it is to get this resolved in LSP protocol in the short term,

I agree.

> This is not ideal of course, but lacking consensus on how to handle it in the LSP itself, we may not have much choice.

I think that in parallel of just making things work here for XML, it's just time to open finer grain issues and maybe even PRs to the LSP spec to handle that.

> BTW: Another option might be to have a 'executeClientCommand' provided by lsp4e

I already don't like a lot the idea of specific operations in Wild Web Developer, but I like much less the idea of placing them in LSP4E ;)
I may seem annoying here, but our goal isn't only to make 1 use-case work, it's also to improve the overall IDE domain and LSP directly. So I'm trying to enforce that before adding what I perceive as a workaround which is bringing back the N*M integration cost; we make a decent proposal in LSP to at least have some chances that this workaround can be replaced by something more standard and cost N+M in the future.

kdvolder:
> I already don't like a lot the idea of specific operations in Wild Web Developer, but I like much less the idea of placing them in LSP4E ;)

I understand, but... in a way putting into lsp4e is one small step closer to making it 'standardized'. So to me this would be better than having every individual Eclipse based client have their own `xml/ExecuteClientCommand`, `sts/executeClientCommand`, `java/executeClientCommand` etc. This is what the current PRs are driving towards. xml is now the third example (that I know of) where we want this kind of thing.

But I totally agree putting non-standard protocol extension in lsp4e vs putting it in xml,  java and spring boot individually, neither is great. The best solution is this be standard part of lsp spec (and then it can be part of lsp4e as well as every other client library that cares to implement the spec faifully).

Anyhow, yes, I think it is probably time to create that LSP ticket to request a protocol extension and see what traction it gets. At least we can write up a bit more carefully what we think this extension should do, and then maybe we implement it in lsp4e accordingly.

I'll try and get round to this in the next couple of days.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/CommandExecutor.java
kdvolder: okay so, PR ws updated. The  CommandExecutor removed and using the lsp4e one now instead.

kdvolder:
I've made the requested change to use CommandExecutor from lsp4e. However, jenkins doesn't like it:

```
[ERROR] /home/jenkins/agent/workspace/Wildwebdeveloper_PR-543/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/XmlLanguageClientImpl.java:[6] 
[ERROR] 	import org.eclipse.lsp4e.command.CommandExecutor;
[ERROR] 	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[ERROR] The import org.eclipse.lsp4e.command.CommandExecutor cannot be resolved
[ERROR] /home/jenkins/agent/workspace/Wildwebdeveloper_PR-543/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/XmlLanguageClientImpl.java:[17] 
[ERROR] 	return CommandExecutor.executeCommand(cmd, null, null);
```

Seems like it isn't there? I beleave that @mickaelistria  did already merge that change into lsp4e. See: https://git.eclipse.org/r/c/lsp4e/lsp4e/+/170859

So... can someone help figure out why the build isn't happy?

mickaelistria:
Thanks Kris.
We'd need to move Wild Web Developer to the snapshots build of LSP4E for this to work. However, it's not really desirable to move to snapshots at the moment since we'd like to release Wild Web Developer soon-ish.
The best path forward seems to be that we 1st release LSP4E with this change; and then upgrade Wild Web Developer to this new LSP4E release so this method becomes visible.
Would that work for you? If so, please send a note to lsp4e-dev mailing-list to request for a release and we'll make it happen unless someone shares good reasons not to.

Note that another possibility would be that you manage to implement support for the `client/executeCommand` according to your LSP proposal directly in LSP4J and get it released soon; then we can work towards support for it in LSP4E, and in your language server and no change would then be required in Wild Web Developer.

kdvolder:
>  However, it's not really desirable to move to snapshots at the moment since we'd like to release Wild Web Developer soon-ish.

Okay, no problem. 

> The best path forward seems to be that we 1st release LSP4E with this change; and then upgrade Wild Web Developer to this new LSP4E release so this method becomes visible.
Would that work for you? 

Sure. I'll post a message to lsp4edev list shortly as you suggest.

> Note that another possibility would be that you manage to implement support for the client/executeCommand according to your LSP proposal directly in LSP4J and get it released soon; then we can work towards support for it in LSP4E, and in your language server and no change would then be required in Wild Web Developer.

I was assuming we keep pursuing the LSP protocol PR independently in parallel. Also I don't see how it would solve the 'we don't want to use snapshot build right now' concern. We'd still need a change to lsp4e to adopt the new standard protocol and so also a new release/snapshot would be needed. I think. So it just adds more steps in between that need to be done before the release.

mickaelistria:
> So it just adds more steps in between that need to be done before the release.

Right, but those steps are more definitive and will be required anyway in a near future; while here is means we do a series of steps for the interim solution, then some steps for the final solution and then some steps to undo the interim. That greatly increase the cost of the feature implementation.
However, it's more up to you, as you're the main one involved in this effort ;)

kdvolder:
Okay, wait, I think it didn't fully understand what you were saying at first. You suggesting we implement the 'client/executeCommand' message directly in lsp4j / lsp4e and the we don't even need this PR. Actually yes that makes sense to me. 

Are you fine with doing it this way even without yet having a 'formal' acceptance of the LSP PR to become officially part of the protocol?

If so, yes, I agree that might actually be the best solution. But if we have to first work through LSP protocol acceptance, then maybe it takes too long and I think we should probably stick with the custom protocol instead.

mickaelistria:
> Are you fine with doing it this way even without yet having a 'formal' acceptance of the LSP PR to become officially part of the protocol?

I'd be fine from my LSP4E-centric perspective; however I think the LSP4J people may have some other (fair) considerations. I think it's more or less their call here.
Is there anything we can help with getting your PR for LSP merged sooner? That would remove a lot of doubt.

mickaelistria (changes requested):
There is a request to release LSP4E soon, with the CommandExecution class made public, so we'll probably merge it before 2020-12 if it still makes sense until then.
In the meantime please ensure that
1. .target file references the snapshots of TM4E (so CommandExecutor is visible)
2. the minimal version for org.eclipse.lsp4e in org.eclipse.wildwebdeveloper.xml is 0.10.0 (which does expose CommandExecutor)

kdvolder:
> the minimal version for org.eclipse.lsp4e in org.eclipse.wildwebdeveloper.xml is 0.10.0 (which does expose CommandExecutor)

Is that correct? Near as I can tell the 'master clone' of lsp4e in my workspace is version 0.13.4. So that should be one where we added 'CommandExecutor' as public api?

kdvolder:
> 1. .target file references the snapshots of TM4E (so CommandExecutor is visible)

I don't understand. Did you mean LSP4E snapshot (instead of TM4E)? Earlier you said we should not depend on the snapshot so now I'm confused.

mickaelistria:
Sorry, you're right, it's LSP4E 0.13.4, I'm lost in too many topics these days... And keep in mind is not really a "public api" but more a temporary exposure of internal stuff until we completed the same story more elegantly ;)

kdvolder:
@mickaelistria I've made the changes I think you wanted but not sure I understood what you wanted (see my confusion / questions above). Please have a look at the changes in last commit and let me know :-)

mickaelistria (changes requested):
Can you please bump the version of the bundle to 0.10.0 ?

mickaelistria (changes requested):
Version in the pom.xml must be matching the version in MANIFEST.MF (except that .qualifier is to be replaced by -SNAPSHOT). You can try tocal builds with `mvn verify` to see such issues more immediately.
It's possible that some other features need similar change as a result of the bundle changing.

kdvolder:
I allways forget to modify the pom on version bumps :-( 

Thanks for the tip on `mvn verify`. Unfortunately it fails with some other error now:

```
ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:2.0.0:package-plugin (default-package-plugin) on project org.eclipse.wildwebdeveloper: /home/kdvolder/git/wildwebdeveloper/org.eclipse.wildwebdeveloper/build.properties: bin.includes value(s) [package-lock.json, node_modules/] do not match any files. -> [Help 1]
```

Not really sure this is an actual problem or not, or just something about running the build locally on my machine. So I'll just push the change adn see what jenkins makes of it.

mickaelistria:
I've merged a commit that's partly conflicting here, adding 1 level of annoyance ;)
I'll take care of the final touch and will merge.

kdvolder:
> I've merged a commit that's partly conflicting here, adding 1 level of annoyance ;)
I'll take care of the final touch and will merge.

I noticed, pushed something to try and address that conflict already.

kdvolder:
There a test fail now. I'd try running this test locally in my runtime workbench, but I'm not sure how to satisfy the dependency on:

```
org.eclipse.ui.tests.harness
```

So at the moment I can't get the test compiling / running.

mickaelistria:
This is merged now. Thanks a lot Alex and Kris!

kdvolder:
@mickaelistria Thanks a lot for shepherding this through :-)

mickaelistria:
Thanks even more to you for keeping this change moving forward despite my numerous change requests!

kdvolder:
Trying to 'consume' this in STS builds but we can't it seems because of:

```
Require-Bundle: 
 org.eclipse.wildwebdeveloper.xml;bundle-version="[0.9.0,0.10.0)",
```

In https://github.com/kdvolder/m2e-core/blob/master/org.eclipse.m2e.editor.lemminx/META-INF/MANIFEST.MF#L13

This is 'pinning' us to version 9.0.0.

PR #560, opened 2020-11-13
Some ls update

mickaelistria:


org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -111,6 +111,7 @@
 							<executable>npm</executable>
 							<arguments>
 								<arg>install</arg>
+								<arg>--only=prod</arg>
```
mickaelistria, line 114: @vrubezhny I found a magic trick to reduce conflicting deps and too long, it makes that devDependencies are ignored and manages to standardize on a single version of estree and so on.

vrubezhny: @mickaelistria  Good. Thanks for that, I'll try again with the current LS updates then

mickaelistria:
Included in #564

PR #593, opened 2021-01-04
Use Eclipse JVM to launch XML LS

### LINKED ISSUE #592: Search for JVM to start XML LS
Currently the JVM search to start XML LS does the following:
1. Check PATH env variable if it points to a JVM
2. Use Java from `System.getProperty("java.home")` which is the Java used to start Eclipse (only if 1 fails)

Why is 1. needed? Should just always be started with whatever JVM launched Eclipse.

In my case (and not just mine) PATH points to Java 8 but Eclipse started with Java 14. XML LS needs Java 11 minimum.

I propose to keep lines 161-165 from here https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/XMLLanguageServer.java#L154 and remove the rest (the PATH variable support)

BoykoAlex:
Fixes #592 

Change-Id: I238f29d94a7ae4f7b731f41dbd289ff8e2b69af6
Signed-off-by: BoykoAlex <aboyko@pivotal.io>

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper.tests/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Tests for WildWebDeveloper
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper.tests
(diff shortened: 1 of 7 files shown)


mickaelistria:
Please bump version of artifacts that are modified (bundle+feature I guess)

mickaelistria:
Thanks for this patch Alex!

Issue #588, opened 2020-12-30
Chrome Debug Fails to Start


samyem:
On Eclipse 2020-12 I have a react project that I am trying to debug with Chrome:
![image](https://user-images.githubusercontent.com/323450/103367147-58c63f00-4a92-11eb-8186-996fc871bc0c.png)

This results in the following error:
```
An internal error occurred during: "Launching chrome debugger".
Cannot invoke "java.io.File.getAbsolutePath()" because the return value of "org.eclipse.wildwebdeveloper.embedder.node.NodeJSManager.which(String)" is null
```
I got the following plugins:
![image](https://user-images.githubusercontent.com/323450/103367322-c70b0180-4a92-11eb-8c09-877909c65a05.png)

mickaelistria:
What OS are you using?
Can you please check in the `<eclipse-workspace>/.metadata/.log` file if there is additional content to help debugging?

samyem:
This is in Windows 10. The only error message in log is:
An internal error occurred during: "Launching chrome debugger".
Cannot invoke "java.io.File.getAbsolutePath()" because the return value of "org.eclipse.wildwebdeveloper.embedder.node.NodeJSManager.which(String)" is null

vrubezhny:
@samyem  Could you still provide the `<eclipse-workspace>/.metadata/.log` or a full stack trace if exist regarding the 'java.io.File.getAbsolutePath()` failure?
Also, could you please check if `Node.js for Windows x86_64` plugin fragment exists in your Eclipse instance? If not, do you have any Node.js version installed on your system?

vrubezhny:
@samyem Do you have Chrome selected in `Browser` tab of your launch configuration?

![chrome-debug-launch-config](https://user-images.githubusercontent.com/620781/104240720-cf811600-545c-11eb-8ede-a3050ef7d8bc.png)

I'm pretty sure in that if you select `Run with: Chrome` then it should normally start debugging.
It looks like it isn't selected by default, so you have to do it manually.
HTH

samyem:
Yes I have Chrome selected.

vrubezhny:
> Yes I have Chrome selected.

It looks like I've got what the problem is. Normally when installing a product on Windows, it's executables aren't added into a PATH if they aren't explicitly needed to be added.
[...]

PR #597, opened 2021-01-11
Chrome Debug Fails to Start #588

vrubezhny:
When Chrome/Firefox Debug launch is created, the only browser defined in 'Run with:` field of launch configuration is to be used to start debuging. So, no any additional searches for a browser executable to be performed.
If no browser is confiured, a message is displayed prompting a user to open Launch Configuration Browser Tab or Web Browser preference page to configure/select a correct browser.
    
Firefox Debug is also updated to serach fora browser executable in a list of browsers defined in Web Browser preference page.
If no executable can be found it is suggested to open Web Browser preference page to add/edit Firefox browser location.

Fixes: #588

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/MessageUtils.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/MessageUtils.java
@@ -1,0 +1,118 @@
+/*******************************************************************************
+ * Copyright (c) 2021 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.wildwebdeveloper.debug;
+
+import java.text.MessageFormat;
+
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.debug.internal.ui.DebugUIPlugin;
+import org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationEditDialog;
+import org.eclipse.debug.internal.ui.launchConfigurations.LaunchGroupExtension;
+import org.eclipse.debug.ui.DebugUITools;
+import org.eclipse.debug.ui.ILaunchConfigurationTab;
+import org.eclipse.debug.ui.ILaunchGroup;
+import org.eclipse.jface.dialogs.IDialogConstants;
+import org.eclipse.jface.dialogs.MessageDialog;
+import org.eclipse.jface.preference.IPreferenceNode;
+import org.eclipse.jface.preference.PreferenceDialog;
+import org.eclipse.jface.preference.PreferenceManager;
+import org.eclipse.jface.preference.PreferenceNode;
+import org.eclipse.jface.window.Window;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.custom.BusyIndicator;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.internal.browser.WebBrowserPreferencePage;
+import org.eclipse.wildwebdeveloper.Activator;
+
+@SuppressWarnings("restriction")
+public class MessageUtils {
+	private static final String BROWSER_TAB_NAME = "Browser"; //$NON-NLS-1$
+	private static final String WEB_BROWSER_PREFERENCE_PAGE = "org.eclipse.ui.browser.preferencePage"; //$NON-NLS-1$
+	
+	private MessageUtils() {
+		// stateless, don't instantiate
+	}
+	
+	/**
+	 * Shows an error message, suggesting to fix browser location either by
+	 * - Selecting a Runtime executable in launch configuration (if applicable)
+	 * - Adding/editing a browser in Web Browser preference page
+	 * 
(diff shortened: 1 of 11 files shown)


mickaelistria:
Code looks good.
I'm wondering whether instead of putting the logic in the LaunchAdapter, we should instead put this logic at bundle startup in order to update the known list of Web Browser directly in the web browser registry, and use the 1st one of them in case of default. What do you think? (note it can be a further iteration, no need to change it as part of this patch unless you think it's worth it)
Can you please bump version of org.eclipse.wildwebdeveloper (and other necessary artifacts) by +0.0.1

akurtakov:
Should https://git.eclipse.org/c/platform/eclipse.platform.ui.git/tree/bundles/org.eclipse.ui.browser/plugin.xml#n164 be changed to be able to find more variations of chrome installs ?

mickaelistria:
> Should https://git.eclipse.org/c/platform/eclipse.platform.ui.git/tree/bundles/org.eclipse.ui.browser/plugin.xml#n164 be changed to be able to find more variations of chrome installs ?

Are those browsers supposed to be added automatically or should user have to do something for them to be discovered? I can't find my firefox configured by default for instance.
I think it would indeed be great if we remove some logic from Wild Web Developer to reuse what's in the browser bundle.

mickaelistria:
Thanks to Alex's hint, I think we should rely more on the browser plugin and even remove the lookup for chrome executables and instead just use the 1st declared browser that matches as default.

vrubezhny:
I agree on that we shouldn't make any browser lookups here, but to use one of the matching  browsers defined on the `Web Browser`preference page. 
The problem is that `Browser` tab is marked as  `OK, No problem` even if no browser is selected as well as in case if a browser selection is done, but path is wrong/empty.
So a user still can try to launch the debugging session (and have the problems described in the issue).

mickaelistria:
So let's have the tab validation turn the tab incomplete if no browser is selected.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/MessageUtils.java
```diff
@@ -0,0 +1,114 @@
...
+		if (suggestEditLaunch) {
+			buttons.add(Messages.RuntimeExecutable_NotDefinedError_OpenLaunchConfigurationBrowserTab);
+		}
+		buttons.add(Messages.RuntimeExecutable_NotDefinedError_OpenWebBrowserPreferencePage);
+		buttons.add(IDialogConstants.CANCEL_LABEL);
+		
+		int result[] = {Window.CANCEL};
+		Display.getDefault().syncExec(() -> {
+			int response = suggestEditLaunch ?
+					MessageDialog.open(MessageDialog.ERROR, Activator.getShell(), title, message, SWT.NONE,
```
mickaelistria, line 66: The shell should be a method parameter as with current implementation, using multiple screens, the dialog can show on the other monitor.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/MessageUtils.java
```diff
@@ -0,0 +1,114 @@
...
+@SuppressWarnings("restriction")
+public class MessageUtils {
+	private static final String BROWSER_TAB_NAME = "Browser"; //$NON-NLS-1$
+	private static final String WEB_BROWSER_PREFERENCE_PAGE = "org.eclipse.ui.browser.preferencePage"; //$NON-NLS-1$
+	
+	private MessageUtils() {
+		// stateless, don't instantiate
+	}
+	
+	public static int showBrowserLocationsConfigurationError(ILaunchConfiguration configuration, String mode, String browserName, boolean suggestEditLaunch) {
```
mickaelistria, line 48: Can you please add a minimal comment to describe what's the return value?

mickaelistria (commented):
Please see minor comments inline.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/MessageUtils.java
vrubezhny: OK

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/MessageUtils.java
vrubezhny: Sure

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/MessageUtils.java
```diff
@@ -0,0 +1,118 @@
...
+									IDialogConstants.CANCEL_LABEL);
+			if (!suggestEditLaunch) {
+				response++;
+			}
+			if (response == 0) {
+				ILaunchGroup group = DebugUITools.getLaunchGroup(configuration, mode);
+				if (group != null) {
+					LaunchGroupExtension groupExt = DebugUIPlugin.getDefault().getLaunchConfigurationManager().getLaunchGroup(group.getIdentifier());
+					if (groupExt != null) {
+						final LaunchConfigurationEditDialog dialog = new LaunchConfigurationEditDialog(Activator.getShell(), configuration, groupExt, false);
```
mickaelistria, line 85: parentShell

vrubezhny: Yes. Sorry, forget about this one

mickaelistria (approved):
+1 to merge this good change as soon as Jenkins validates it.
It's a very good improvement from code and UX perspective, relying solely on upstream standard configuration!

mickaelistria:
Build on Jenkins is passing well this time, let's merge it now and we'll deal with the occasional mirror issue later.

PR #596, opened 2021-01-07
Update Embedded Node.js Runtime to latest LTS version (v14.15.4)

vrubezhny:
Signed-off-by: Victor Rubezhny <vrubezhny@redhat.com>

PR #610, opened 2021-02-02
Update to Fedora 33 docker image

akurtakov:
This means Node.js switches from 12 to 14 and we would better run our
tests with latest LTS version.


akurtakov:
This should go in with #596 but it involves @mickaelistria building and uploading new image.

PR #623, opened 2021-02-13
Resolve URLs used in org.eclipse.wildwebdeveloper.json.schema extensi…

### LINKED ISSUE #622: Handling special schemas in org.eclipse.wildwebdeveloper.json.schema
The `org.eclipse.wildwebdeveloper.json.schema` extension points allows to associate an URL with a pattern.
As this is communicated to the LS, my clain is that only external URLs (http,..) are allowed here.
I think it would be good if we allow special URLs here:

- platform:
 - plugin:

That would allow plugin provider to bundle the schema inside the plugin
We should document clearly which schemes are supported

jeffmaury:
…on point

Fixes #622

Signed-off-by: Jeff MAURY <jmaury@redhat.com>

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/build.properties
+++ b/org.eclipse.wildwebdeveloper.tests/build.properties
@@ -1,4 +1,5 @@
-source.. = src/
+source.. = src/,\
+           resources/
 output.. = bin/
 bin.includes = META-INF/,\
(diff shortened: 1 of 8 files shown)


mickaelistria:
Thanks for this good addition!

PR #626, opened 2021-02-15
Implemented "openDoc" to open the given url in an External browser

jcompagner:
This adds the "openDoc" implementation of ESLint possible quickfixes

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/eslint/ESLintClientImpl.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/eslint/ESLintClientImpl.java
@@ -12,15 +12,22 @@ package org.eclipse.wildwebdeveloper.eslint;
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URL;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.lsp4e.LanguageClientImpl;
 import org.eclipse.lsp4j.ConfigurationItem;
 import org.eclipse.lsp4j.ConfigurationParams;
+import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.browser.IWorkbenchBrowserSupport;
+import org.eclipse.wildwebdeveloper.Activator;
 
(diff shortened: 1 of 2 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/eslint/ESLintClientImpl.java
```diff
@@ -75,4 +78,17 @@
...
 		return CompletableFuture.completedFuture(null);
 	}
+	
+	@SuppressWarnings("unchecked")
+	@Override
+	public CompletableFuture<Void> openDoc(Object o) {
+		if (o instanceof Map && ((Map<?, ?>) o).containsKey("url")) {
+			IWorkbenchBrowserSupport browserSupport = PlatformUI.getWorkbench().getBrowserSupport();
+			try {
+				browserSupport.getExternalBrowser().openURL(new URL(((Map<String, String>) o).get("url")));
```
mickaelistria, line 88: We should honor user preferences and not force usage of external browser if users has preference set to prefer internal browser. So `browserSupport.createBrowser(...)` should be used instead.

jcompagner: ok i changed to that,
its currently kind of useless under windows because of the default IE browser that is still there
(and i think many of those doc stuff go to github and github doesnt support IE for quite a while now)

hopefully this one https://bugs.eclipse.org/bugs/show_bug.cgi?id=538991 and/or the chromium port is getting more mature and default as soon as possible.

akurtakov: Setting up chromium is too much of a trouble and not shipped by default with swt due to its HUGE size thus when there is a change in default it would be edge and not chromium

mickaelistria:
Thanks, I'm merging it as the previous version was already good (I tested it), new code has fixed only concern and the failures are more infra problems.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/eslint/ESLintClientImpl.java
jcompagner: right and one of the 2 is fine for this scenario
Our product already ships with that chromium, because we heavily depend on the internal browsers for some editors and we want to be platform independent, so we like to have the same browser everywhere, but now with Apple M1 thats also tricky,,

And i am not saying WWD needs to do anything for that ;) 
It was just a comment of my side that using the internal browser is less and less useful currently in the windows world
(but also mac is not handy, we also have problems with the embedded Safari)

PR #618, opened 2021-02-10
VSCode 1.54

mickaelistria:
Signed-off-by: Mickael Istria <mistria@redhat.com>

PR #628, opened 2021-02-15
upgraded vscode to the latest release build

jcompagner:


jcompagner:
ah i see there was already a pull for this: https://github.com/eclipse/wildwebdeveloper/pull/618

and that there was a problem with that (i didn't really notice any problem, but i guess i don't really touch that code)

jcompagner:
aargh

i didn't see it because that stupid download-maven-plugin wget plugin is caching to much

Don't you guys have this problem? i really dont' get that.
It seems to cache based on the target (so what file you download to)  instead of just the url.

thats why i did <skipCache>true</skipCache> with the eslint package.json wget ...

mickaelistria:
Please report ossue about download-maven-plugin to download-maven-plugin.
I'm closing this one as it's a duplicate of #618

jcompagner:
i get this is a problem of the maven plugin :)
i just wondered if you guys never had this problem. stupid thing is i really checked the caching code and i don't see a big problem there.. They really cache it by the url, but for me it seems to be cached by the target name..  I had this problem before in other projects and i just make sure the target name is also always different. But why i have that i have no idea.

PR #629, opened 2021-02-15
added language-service back in an updated angular stuff to 11.2

jcompagner:


mickaelistria:
@angular/language-service is coming as a transitive dependency.

PR #632, opened 2021-02-18
clean node_modules as a whole and only the package-lock.json in the root dir

jcompagner:
Signed-off-by: Johan Compagner <jcompagner@gmail.com>


mickaelistria:
Thanks!

PR #637, opened 2021-02-23
XML CodeLens preference page

### LINKED ISSUE #636: XML CodeLens preference page
Add a XML Codelens preference page to benefit with codelens inside XSD and DTD. See https://github.com/redhat-developer/vscode-xml/blob/master/docs/CodeLens.md

angelozerr:
XML CodeLens preference page

Fixes #636

### DIFF
--- a/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/ui/messages.properties
+++ b/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/ui/messages.properties
@@ -11,7 +11,7 @@
 # *******************************************************************************/
 
 # Preference page
-XMLPreferencesPage_XMLCatalogsLink=See <A>'XML Catalogs'</A> for XML catalogs preferences
+XMLPreferencePage_XMLCatalogsLink=See <A>'XML Catalogs'</A> for XML catalogs preferences
 
 # XML Catalog preference page
 XMLCatalogPreferencePage_Entries=XML Catalog Entries:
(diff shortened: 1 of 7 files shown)


angelozerr:
With this PR you can enable XMl CodeLens inside XSD and DTD:

image

The click on code lens doesn't work because it requires to implement the proper command to execute the references. But we could support that in an another PR?

mickaelistria:
I think it would be better to have an "Edition" preference page with the codelens entry.

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/ui/messages.properties
```diff
@@ -21,6 +21,9 @@ XMLCatalogPreferencePage_OpenInEditorMessage=Do you want to Apply the changes an
 XMLCatalogPreferencePage_OpenInEditorApplyAndEdit=Apply and Edit
 XMLCatalogPreferencePage_OpenInEditorNo=No
 
+# XML CodeLens preference page
+XMLCodelensPreferencePage_codelens_enabled=Enable/disable XML &CodeLens
```
mickaelistria, line 25: Just "Enable XML Codelens" as it's a checkbox, ideally also with a description of what are the expected codelens (reference count on schemas)

angelozerr:
> I think it would be better to have an "Edition" preference page with the codelens entry.

In the future I would like to provide codelens settings per codelens kind (ex : xml.codeLens.schema.enabled,  xml.codeLens.dtd.enabled, etc), so I think it's better to have a Codelens page for that (to be consistent with JDT we have a CodeMining page).

org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/ui/messages.properties
angelozerr: > Just "Enable XML Codelens" as it's a checkbox, ideally also with a description of what are the expected codelens (reference count on schemas)

LemMinx provides this feature but LemMinx extension could provide Codelens too (ex : Codelens to execute a goal for maven), so it's not only restricted for XSD and DTD.

mickaelistria: OK. It's more the "Enable/Disable" part which is not good UI workding besides a checkbox.

mickaelistria:
Merged as f93fd6810a32e22a7e5e1218fb5135668ae67bd6 . Thanks!

PR #643, opened 2021-02-28
implemented a LanguageClientImpl for angular

jcompagner:
added json request/notifications for:
projectLoadingStart, projectLoadingFinish, projectLanguageService,
suggestIvyLanguageServiceMode, suggestStrictMode and implemented
notifyProgress

For now these all just log, but suggestIvyLanguageServiceMode should
give the user an option (or point to an option) to enable the
commands.add("--experimental-ivy");
in the AngularLanguageServer

(this should i think be a setting in the preferences)
Problem is a bit if you enabled this now that you get many exception in
the log because of: https://github.com/eclipse/lsp4j/issues/479

Signed-off-by: Johan Compagner <jcompagner@gmail.com>

### DIFF
--- a/org.eclipse.wildwebdeveloper.feature/feature.xml
+++ b/org.eclipse.wildwebdeveloper.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.feature"
       label="%name"
-      version="0.10.6.qualifier"
+      version="0.10.7.qualifier"
       provider-name="Eclipse Wild Web Developer project"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
--- a/org.eclipse.wildwebdeveloper.feature/pom.xml
+++ b/org.eclipse.wildwebdeveloper.feature/pom.xml
@@ -7,7 +7,7 @@
 		<version>0.5.0-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-feature</packaging>
-	<version>0.10.6-SNAPSHOT</version>
+	<version>0.10.7-SNAPSHOT</version>
 	<build>
 		<plugins>
 			<plugin>
--- a/org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: Wild Web Developer: web development in Eclipse IDE
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper;singleton:=true
 Automatic-Module-Name: org.eclipse.wildwebdeveloper
-Bundle-Version: 0.5.14.qualifier
+Bundle-Version: 0.5.15.qualifier
 Bundle-Activator: org.eclipse.wildwebdeveloper.Activator
 Bundle-Vendor: Eclipse Wild Web Developer project
 Require-Bundle: org.eclipse.ui,
--- a/org.eclipse.wildwebdeveloper/pom.xml
+++ b/org.eclipse.wildwebdeveloper/pom.xml
@@ -7,7 +7,7 @@
 		<version>0.5.0-SNAPSHOT</version>
 	</parent>
 	<packaging>eclipse-plugin</packaging>
-	<version>0.5.14-SNAPSHOT</version>
+	<version>0.5.15-SNAPSHOT</version>
 
 	<build>
 		<plugins>
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/angular/AngularClientImpl.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/angular/AngularClientImpl.java
@@ -1,0 +1,53 @@
+package org.eclipse.wildwebdeveloper.angular;
+
+import java.util.Map;
+
+import org.eclipse.lsp4e.LanguageClientImpl;
+import org.eclipse.lsp4j.MessageParams;
+import org.eclipse.lsp4j.MessageType;
+import org.eclipse.lsp4j.ProgressParams;
+import org.eclipse.lsp4j.WorkDoneProgressNotification;
+import org.eclipse.lsp4j.WorkDoneProgressReport;
+
+public class AngularClientImpl extends LanguageClientImpl implements AngularLanguageServerExtention {
+
+	@Override
+	public void projectLoadingFinish(Object object) {
+		// TODO should this set some state because only now stuff will work like hover..
+		// or maybe even after projectLanguageService "enabled" call
+		logMessage(new MessageParams(MessageType.Info, "Angular project loading finished"));
+	}
+	
+	@Override
+	public void projectLoadingStart(Object object) {
+		logMessage(new MessageParams(MessageType.Info, "Angular project loading started"));
+	}
+	
+	@Override
+	public void projectLanguageService(Map<String,Object> data) {
+		logMessage(new MessageParams(MessageType.Info, "Language Service is " + (((Boolean)data.get("languageServiceEnabled")).booleanValue()?"":"not yet ") + "enabled for project " + data.get("projectName")));
+	}
+
+	@Override
+	public void suggestIvyLanguageServiceMode(Object o) {
+		logMessage(new MessageParams(MessageType.Info, o.toString()));
+		// TODO should this propose a setting (in the preferences) that enables --experimental-ivy command in the AngularLanguageServer)
+	}
+
+	@Override
+	public void suggestStrictMode(Object o) {
+		// this only says to the developer that they should enabled something in there tsconfig file (strictTemplates: true)
+		// can't do any from there (like  suggestIvyLanguageServiceMode has to do)
+		logMessage(new MessageParams(MessageType.Info, o.toString()));
+	}
+	
+	@Override
+	public void notifyProgress(ProgressParams params) {
+		String message = null;
+		WorkDoneProgressNotification value = params.getValue();
+		if (value instanceof WorkDoneProgressReport) {
+			message = ((WorkDoneProgressReport) value).getMessage();
+		}
+		logMessage(new MessageParams(MessageType.Info, params.getToken().getLeft()  + ": " + (message == null? "done":message)));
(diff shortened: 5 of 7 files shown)


mickaelistria:
Please bump version by +0.0.1.
About Ivy, I'm not found of adding preferences that user will not really get clue about their existence or utility. The main question is when Ivy should be enabled or not, and whether we can automate that decision. Or, even better, do nothing and just use Ivy when the Angular LS uses it as default. After all, there are more reasons to trust Angular LS decision than to make it a user decision.

jcompagner:
The thing is that Visual Studo code has a preference for that, because this is really a client side thing (you need to add that argument)

Problem is that i as a user did decide that i want to use strictTemplates. i have that in the tsconfig.json
And the compiler (ng build) does then find a lot more (potential) problems
But if i don't feed that into the angular service from the user of the angular service i still use the ViewEngine not Ivy.. So that doesn't report all those things. so you need to really look in the console of ng build to get all kind of problems and then go into the editor and fix those and you don't see that...

But yes in the end Angular i guess will switch to be that as default.. It would be just nice if this would be configurable in WWD (even if it is quite hidden away). I now just use it because i run my own eclipse always in debug mode through another eclipse with WWD as source. so i just change the source itself to force WWD to send that IVY command.

But sending that IVY command in a plain WWD without also fixing LSP4J (and because of the upgrade also fix stuff in LSP4E because of api changes in LSP4J)  a lot of errors will be filling up the log because of that LSP4J problem

But anyway this commit doesn't do anything related to that, thats just adding all the angular client notifications so they are logged nicely and you don't get constantly all those errors of not supported json notification/request.

If we can think about so nicer implementations that would be nice
For example i still get a lot of LSPHover timeout exceptions, because the hover implementation already starts working long before the Angular Service is really up and running... I think if you ask the system for all active LangugageServices (what hover code does) we should not give back the angular service yet until projectLoadingFinish is called or even projectLanguageService with a true. Because until then its not there yet...

mickaelistria:
Isn't it a Language Server limitation that this flag must be explicitly passed at startup and the Ivy/not-Ivy switch is not taking tsconfig into account? We usually expect that LS do infer as much as possible about the project according to project metadata; preferences are less flexible and are contrary to that. So they add maintenance cost/technical debt to Wild Web Developer for a relatively bad solution; I'd rather avoid this kind of approaches in WIld Web Developer and focus on making the case by a fix upstream, in the language server.
Anyway, I suggest we first wait for LSP4J to have this bug fixed before merging such change.

jcompagner:
this is how it is designed see: https://blog.angular.io/under-the-hood-of-the-language-service-ab763c26f522

This is because since i think angular 11 they had a lot of changes in the LS code and they are slowly now switching the LS code to also use Ivy. But this is for now experimental (but for me it works nicely with my patches)

without that flag, that you can set in VisualStudioCode, Angular LS still uses ViewEngine not Ivy even if you have strictTemplates to true (and use Ivy fully with ng build)

I think for them those are 2 different worlds. with different roadmaps.

I guess maybe with Angular 12 or 13 that becomes default anyway, not sure what the roadmap for that is.
But if it becomes default then LSF4J must have fixed its problem with the notification parsing else a lot of exceptions will be filling the logs..

mickaelistria:
See build log, feature bump is incomplete.

jcompagner:
phhhh.. i always forget that.. (would be nice if Mickael would fix this so its automatically done ;) )

mickaelistria:
If you have a baseline configured in your IDE, and use the m2e Tycho connector, I think you should see those errors in the IDE with quick-fixes. But it require some configuration effort.
At build-time, build shouldn't change the source code without user explicitly asking for it; it's a good time to fail, not to automate fixes.

mickaelistria:
Note that running `mvn verify -DskipTests` locally should show you such errors more immediately.

jcompagner:
> Note that running `mvn verify -DskipTests` locally should show you such errors more immediately.

hmm but then you get into this:

ERROR] Failed to execute goal org.eclipse.tycho:tycho-packaging-plugin:2.2.0:build-qualifier-aggregator (default-build-qualifier-aggregator) on project repository: Working tree is dirty.
[ERROR] git status repository:
[ERROR]
[ERROR] Changes not staged for commit:
[ERROR]         modified:    repository/epp.product
[ERROR]
[ERROR] You are trying to use tycho-buildtimestamp-jgit on a directory that has uncommitted changes (see details above).
[ERROR] Either commit all changes/add files to .gitignore, or enable fallback to default timestamp provider by configuring
[ERROR] jgit.dirtyWorkingTree=warning for tycho-packaging-plugin

which is for me very annoying... 
what is the real reason for this? why is it blocked that i must first commit before i can try a build? 
and i think by default you can't also really override it by a property that you give at the command line i think for that you need to configure this?

mickaelistria:
See https://wiki.eclipse.org/Tycho/Reproducible_Version_Qualifiers#What_does_this_actually_do.3F about why commit matters.

jcompagner:
https://github.com/angular/angular/commit/e9e7c33f3c170648ec8c86d859980c7fe78fba39

They make it now configurable from the 'other' side...

jcompagner:
> See https://wiki.eclipse.org/Tycho/Reproducible_Version_Qualifiers#What_does_this_actually_do.3F about why commit matters.

but still i don't care if i want to quickly verify/build my changes locally.
as far as i see now that is not controllable by a property that you give to the mvn command i need to add <jgit.dirtyWorkingTree>warning</jgit.dirtyWorkingTree>

which i already added to a few pom files of wwd.. 
I think you need to specify a property? and then that property can be overwritten by a mvn command line argument? if i could just do

mvn clean verify -DdirtyWorkingTree=warning

and be done with it it was fine by me, but as far as i can see that is not directly possible.
so now i always have the annoying fact that i have pom file changes..

mickaelistria:
A PR that sets the `<jgit.dirtyWorkingTree>${jgit.dirtyWorkingTree}</jgit.dirtyWorkingTree>` in the pom so the value can easily be overriden via command-line would be welcome

jcompagner:
are you also planning to include this one?
This is not about that enabling of experimental api
but just a first setup to also get rid of all those warnings in the log just look for example here:

https://ci.eclipse.org/wildwebdeveloper/job/Wildwebdeveloper/job/PR-642/6/console

and search for "angular/"

mickaelistria:
To be honest, the fact that there is the bug in LSP4J that would make turning on the option kind of a bug and the fact that Angular LS is not yet 100% clear on what should be the default between Ivy and non-Ivy, and even the fact that it has to be a flag when starting the LS that some data inferred from the project; all that gives me the impression it's a bit early for WIld Web Developer to expose the option.
What's also stupid with Angular LS is what about a workspace where there are some related Ivy and non-Ivy projects? Should mulitple LS instances be started? That would be a pity, especially if projects share components.
I have the impression it would be better to just fully switch to Ivy and Angular LS and LSP4J are ready for it than offering the option now.
What do you think?

jcompagner:
but this case is not about exposing any options..
I agree fully that if if LSP4J doesn't fix first it stull (and LSP4E on top of that again also needs fixes, i have patches in both) it doesn't make much sense to have an option to add this to the AngularLanguageServer commands.add("--experimental-ivy"); 

which i now do locally.

This patch is nothing more then the eslint_nolibrary it just introduces a Client part that at least handles the stuff angular now sends anyway...

if we wouldnt do this patch but we would add that flag, then the result would just be even more errors and warnings in the log.

In the end we do need a AngularClientImpl anyway i think to handle the stuff angular sends over

i think LSF4J or E already has something that says getLanguageServers() and they filter on when they are ready
I think (not sure how to bind that) that only the client part can tell the server part that it is ready..

I get loads of hoover timeout exceptions in my log files.. this is because it takes a while for the language server to be fully up and running and currently LSF4J already reports it up and running when it is not done yet (the projectLoadingFinish of the angular client still needs to happen)

mickaelistria:
Sorry, I was thinking about other things/patches I've seen. Indeed, this patch is good to add. I'm retriggering the builds hoping the test failures are just "bad luck" and if everything is fine, we can merge.

jcompagner:
thx, i think the more tricky part is when angular maybe in angular 12 oe 13? who knows
makes also the language servers just follow the strictTemplates
so if i have enabled that strictTemplates (which we do in our project) then the language servers automatically already switches over to ivy mode (so we don't need that flag in our server portion) then a lot of exceptions will be filling the logs because LSF4J doesn't implement yet the more generic notification message, which angular then sends over a lot.

so i do hope that in the next release cycle this case https://github.com/eclipse/lsp4j/issues/479 is fixed in LSF4J and that LSF4E is adjusted to the new api and that WWD builds then on top of both 

(how are these release cycles now anyway i wonder? because you really need first a LSF4J then an LSF4E then WWD can build on top of that...)
i think you released a LSF4E not so long ago right? but does WWD already uses that release or is that always coming 1 step behind?

mickaelistria:
> i think you released a LSF4E not so long ago right? but does WWD already uses that release or is that always coming 1 step behind?

Thanks to OSGi, the binding is more flexible: we don't need to rebuild LSP4E or Wild Web Developer for them to use newer LSP4J if present. As long as OSGi requirement matches and no API is broken, users would just have to install newer LSP4J on their machine to have the fix, without need to install newer Wild Web Developer nor LSP4E.
For Wild Web Developer, as this will be a constraint for proper development; we can and should express this constraint in the MANIFEST.MF so newer LSP4J will be installed together with newer Wild Web Developer. We don't need to alter LSP4E for that case, unless newer release of LSP4J breaks API used by LSP4E.

jcompagner:
but lsp4e has hard version requirements 

 org.eclipse.lsp4j;bundle-version="[0.10.0,0.11.0)",
 org.eclipse.lsp4j.jsonrpc;bundle-version="[0.10.0,0.11.0)",

(i need to change that now i have both just checked out, running from source (master branch))

And in the LSP4J 0.11 version it did break api, so LSP4E needs to be adjusted:

its the api change for this:
TextEdit textEdit = item.getTextEdit(); (item == CompletionItem)
which now returns:

Either<TextEdit, InsertReplaceEdit> either = item.getTextEdit();

mickaelistria:
OK, so we'd indeed need to change LSP4E and release it. But I think once LSP4E is fixed and available, we wouldn't need to change Wild Web Developer for the change to take effect.

mickaelistria:
Merged, thanks!

Issue #680, opened 2021-06-07
Refactoring throws fatal error


gbeneke:
Refactoring does not work for me using even the simplest JS code snippets opened in the generic editor.  The error message shown is

![image](https://user-images.githubusercontent.com/85511748/121061647-66f85400-c7c4-11eb-8b84-7ed0e8e9d992.png)

For preview the same problem applies.  Finding references works and finds all variable references that need the refactoring applied.

Screenshot was taken testing the WWD supplied IDE on https://download.eclipse.org/wildwebdeveloper/releases/latest/IDEs/ .  Test was run on Fedora 34 (fresh installation just for this purpose on VirtualBox 6.1.22) but same result applies to Windows 10 (1809) native and with WSL 1 (Ubuntu 20.04) and Ubuntu 20.10.  On all platforms I run into the same problem.  Error messages differ depending on the JDK used.

Using the JEE Eclipse package which comes with WWD 0.10.6 prepackaged shows a different error message (in this case I use the OpenJDK 11.0.11 SDK to drive Eclipse):

![image](https://user-images.githubusercontent.com/85511748/121063843-1fbf9280-c7c7-11eb-9b98-abcd6fb11c17.png)

Error Log view:

![image](https://user-images.githubusercontent.com/85511748/121065016-94470100-c7c8-11eb-82a4-ae5a32394afb.png)

Complete stack trace in the Error Log view:

java.lang.reflect.InvocationTargetException
	at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:397)
	at org.eclipse.ltk.internal.ui.refactoring.RefactoringWizardDialog2.run(RefactoringWizardDialog2.java:324)
	at org.eclipse.ltk.ui.refactoring.RefactoringWizard.internalPerformFinish(RefactoringWizard.java:636)
	at org.eclipse.ltk.ui.refactoring.UserInputWizardPage.performFinish(UserInputWizardPage.java:148)
	at org.eclipse.lsp4e.operations.rename.LSPRenameRefactoringWizard$RenameInputWizardPage.performFinish(LSPRenameRefactoringWizard.java:98)
	at org.eclipse.ltk.ui.refactoring.RefactoringWizard.performFinish(RefactoringWizard.java:710)
	at org.eclipse.ltk.internal.ui.refactoring.RefactoringWizardDialog2.okPressed(RefactoringWizardDialog2.java:450)
[...]

PR #691, opened 2021-06-29
[tests] Refactoring throws fatal error #680

vrubezhny:
A JUnit test is added for the rename issue

Issue: #680 


vrubezhny:
Currently on Jenkins there is exactly one (this one) test failing, so I believe it's OK to be merged.

mickaelistria:
Since this bug is not critical, it's not worth polluting the CI output for other work, Let's keep the test in this branch and make sure we stack other related patches on top of it, and we'll merge it when we have the solution.

PR #698, opened 2021-07-02
[tests] Refactoring throws fatal error #680

vrubezhny:
Fixed TestJsTs.testRefactoringRename() to make it to give more stable results

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestJsTs.java
+++ b/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestJsTs.java
@@ -134,13 +134,14 @@ public class TestJsTs {
 				protected boolean condition() {
 					return renameDialogOkPressed.get();
 				}
-			}.waitForCondition(display, 1000), "Rename dialog not shown");
+			}.waitForCondition(display, 2000), "Rename dialog not shown");
+
 			assertTrue(new DisplayHelper() {
 				@Override
 				protected boolean condition() {
 					return newContent.equals(document.get());
 				}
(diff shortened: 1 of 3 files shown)


org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestJsTs.java
```diff
@@ -135,6 +135,10 @@ protected boolean condition() {
 					return renameDialogOkPressed.get();
 				}
 			}.waitForCondition(display, 1000), "Rename dialog not shown");
+
+			DisplayHelper.sleep(2000); // Give time for LS to initialize enough before making edit and sending a
```
mickaelistria, line 139: Would adding 2000ms to the next usage of DisplayHelper be enough?
Usually, we should avoid sleep, but instead find a good condition to continue and use waitForCondition()

vrubezhny: I've tested the increasing the timeout for waitCondition() - it doesn't help, What I've seen is that rename request was made, successful response is received with the correct location array in place, but nothing has changed on editor until it's closed. Looks like some listeners weren't invoked to update the editor input for some reason or some thread is blocked while we're waiting for the condition.

vrubezhny: I've removed `DisplayHelper.sleep(2000)` line and increased the timeout for waitCondition() to 5 sec - it looks like a minimal timeout value which allow the test to work more or less stable. Visually the rename wizard take some time to appear and make all the required request before the required changes can really be performed on the document, and we should take these operations duration into account.

PR #706, opened 2021-08-06
updates to all the npm packages

jcompagner:
including angular 12.2 but this does mean that we need to be on the
latest lsp4j/lsp4e because of the notification protocol.


mickaelistria:
cc @vrubezhny who is planning to spend time on upgrading language servers soon.

mickaelistria:
Merged as part of #713

PR #707, opened 2021-08-06
Use new lsp4j notification api

jcompagner:
lsp4j now uses has Either support and the Progress now works

### DIFF
--- a/org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
@@ -8,9 +8,9 @@ Bundle-Activator: org.eclipse.wildwebdeveloper.Activator
 Bundle-Vendor: Eclipse Wild Web Developer project
 Require-Bundle: org.eclipse.ui,
  org.eclipse.core.runtime,
- org.eclipse.lsp4e;bundle-version="0.5.0",
- org.eclipse.lsp4j.jsonrpc;bundle-version="0.5.0",
- org.eclipse.lsp4j;bundle-version="0.5.0",
+ org.eclipse.lsp4e;bundle-version="0.13.6",
+ org.eclipse.lsp4j.jsonrpc;bundle-version="0.12.0",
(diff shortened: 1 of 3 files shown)


mickaelistria:
The target-platform needs to be updated to include the newer LSP4E version; and MANIFEST.MF reference newer LSP4J version as well.

jcompagner:
> The target-platform needs to be updated to include the newer LSP4E version; and MANIFEST.MF reference newer LSP4J version as well.

But it is not released yet right (LSP4E) ? So i can do the manifest but the target-platform is i think not really possible yet?

I was just pushing this to be ready for the release of LSP4E

mickaelistria:
> But it is not released yet right (LSP4E) ?

No, but you can reference the snapshots site in the .target.

mickaelistria:
A new LSP4E release just happened; please update your PR to use it.

jcompagner:
but the target file is with version 0.0.0 anyway: https://github.com/eclipse/wildwebdeveloper/blob/master/target-platform/target-platform.target#L41

so those auto map right?

should i then change the minimum version in the feature?
currently that has 0.5.0 for lsp4j and lsp4e:

https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF#L11

i guess we need to set that for lsp4j to 0.12 and lsp4e to 0.13.6 ?

mickaelistria:
> i guess we need to set that for lsp4j to 0.12 and lsp4e to 0.13.6 

yes please

mickaelistria:
Can you please isolate the patch about support for notifications here; and keep the LS upgrade separated? We could merge this one as soon as test are passing, but we'll need to go through CQ and usually more fixes when upgrading Language Servers.

jcompagner:
i made another pull request that should have the parent from master.

mickaelistria:
Thanks. In such case, you can use `git push --force` to your remote branch in order to replace the content of the PR. With that, you don't need to create a new PR.

PR #713, opened 2021-08-31
Update npm packages 

vrubezhny:
Update NPM modules to the following versions:

typescript@4.3.5
typescript-language-server@0.5.4
typescript-plugin-css-modules@3.4.0
firefox-debugadapter@2.9.4

yaml-language-server@0.19.2
@angular/language-server@11.2.14

### DIFF
--- a/org.eclipse.wildwebdeveloper/pom.xml
+++ b/org.eclipse.wildwebdeveloper/pom.xml
@@ -144,18 +144,18 @@
 								<arg>--no-bin-links</arg>
 								<arg>--ignore-scripts</arg>
 								<arg>${project.build.directory}/node-debug2-1.42.5.tgz</arg>
-								<arg>typescript@4.1.3</arg>
-								<arg>typescript-language-server@0.5.1</arg>
+								<arg>typescript@4.3.5</arg>
+								<arg>typescript-language-server@0.5.4</arg>
 								<arg>typescript-lit-html-plugin@0.9.0</arg>
-								<arg>typescript-plugin-css-modules@3.1.0</arg>
+								<arg>typescript-plugin-css-modules@3.4.0</arg>
 								<arg>${project.build.directory}/vscode-css-languageserver-1.0.0.tgz</arg>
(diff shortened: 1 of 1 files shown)


mickaelistria:
Does `testRefactoringRename` work for you locally? It's been repeatedly failing on CI, maybe the text needs a tweak.

vrubezhny:
> Does `testRefactoringRename` work for you locally? It's been repeatedly failing on CI, maybe the text needs a tweak.

The rename refactoring itself works fine, while the test locally fails in most of the runs. Tweaking it with increasing timeout values doesn't help... I'm going to look more deeply if any wizard's UI elements/labels have changed - the test is very sensitive to such changes.

mickaelistria:
I submitted https://dev.eclipse.org/ipzilla/show_bug.cgi?id=23711 .
Are there any of those npm deps that are not using latest version? If so, can you please details the issue that happen using latest releases?

mickaelistria:
I see yaml and angular LS have been using lower version. Can you please add as comment in the pom.xml the failure that are caused by latest releases?

vrubezhny:
@mickaelistria Thanks for creating the CQ!
Yes YAMS and Angular LSs are not of the latest versions. Using the latest versions for these two makes their according tests to fail - I didn't spot any obvious exceptions or errors, but it looks like they fail due to no diagnostics messages received.
I'll comment on pom.xml on failing tests for these LSs.

mickaelistria:
CQ was approved, let's merge.

vrubezhny:
@mickaelistria  Should we close the original PR from @jcompagner  
- https://github.com/eclipse/wildwebdeveloper/pull/706 or we can still reuse it later for updating Angular and YAM LSs?

jcompagner:
Is there a reason why we only update to 11.x now that 12.x is already out for quite some time (and I think 12 has some nice additions)

Or did that introduce that go-to stuff that doesn't really nicely work (as mentioned in another case)

PR #727, opened 2021-09-16
Fix Rename JS-TS Identifier Test

vrubezhny:
Fix Rename JS-TS Identifier Test

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestJsTs.java
+++ b/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestJsTs.java
@@ -16,6 +16,8 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.ByteArrayInputStream;
 import java.lang.reflect.Method;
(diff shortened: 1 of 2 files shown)


org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
```diff
@@ -8,7 +8,7 @@ Bundle-Activator: org.eclipse.wildwebdeveloper.Activator
 Bundle-Vendor: Eclipse Wild Web Developer project
 Require-Bundle: org.eclipse.ui,
  org.eclipse.core.runtime,
- org.eclipse.lsp4e;bundle-version="0.13.6",
+ org.eclipse.lsp4e;bundle-version="0.13.8",
```
mickaelistria, line 11: Why is this necessary? If not, let's just leave it as it since 0.13.8 is not in SimRel and causes the build to fail.

vrubezhny: Sorry. This is a trace left from using a locally built version of LSP4E. No need to be here. Fixed

mickaelistria:
Thanks!

PR #729, opened 2021-09-29
Update build system

akurtakov:


mickaelistria:
Changing the bundle version if there is no payload change means we'll trigger all existing users to download dozens (if not hundreds) of extra megabytes when updating, for no additional valuento accomodate the cost. That's not very green-it.

akurtakov:
All the bundle bumps were required as local build failed without them. It's quite likely that this happens due to different npm/node versions on build system and locally but I can't think how to fix that.

mickaelistria:
The need to bump version in all bundles although there is no payload change is most likely caused by the bump of the version in parent and modification of all children pom. pom.xml are *not* filtered out when computing timestamp, since a lot of the content of the artifact (all language servers) actually comes from the pom definition.

akurtakov:
Ah, you're probably correct. I wanted to use the parent pom as a definition for the version but looks like this is not easily achievable here.

akurtakov:
This PR is just maven plugins bumps now.

PR #738, opened 2021-10-29
Support .cjs and .mjs JavaScript extension variants

PyvesB:
`.cjs` and `.mjs` are two variants of the standard `.js` JavaScript file extension, which can be used to differentiate between the CommonJS and ESM module systems. More information can be found in the [Node.js documentation](https://nodejs.org/docs/latest/api/packages.html#determining-module-system).


jabby (approved):

mickaelistria:
Thanks

PR #749, opened 2021-11-25
Update to the latest lsp4e (0.13.8) version

jcompagner:
to get the fix:
https://github.com/eclipse/lsp4e/issues/3

### DIFF
--- a/org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper/META-INF/MANIFEST.MF
@@ -3,12 +3,12 @@ Bundle-ManifestVersion: 2
 Bundle-Name: Wild Web Developer: web development in Eclipse IDE
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper;singleton:=true
 Automatic-Module-Name: org.eclipse.wildwebdeveloper
-Bundle-Version: 0.5.19.qualifier
+Bundle-Version: 0.5.20.qualifier
 Bundle-Activator: org.eclipse.wildwebdeveloper.Activator
 Bundle-Vendor: Eclipse Wild Web Developer project
 Require-Bundle: org.eclipse.ui,
  org.eclipse.core.runtime,
- org.eclipse.lsp4e;bundle-version="0.13.6",
+ org.eclipse.lsp4e;bundle-version="0.13.8",
  org.eclipse.lsp4j.jsonrpc;bundle-version="0.12.0",
  org.eclipse.lsp4j;bundle-version="0.12.0",
  org.eclipse.tm4e.registry;bundle-version="0.1.0",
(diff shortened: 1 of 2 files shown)


mickaelistria:
Thanks!
Note that the requirement here is a version range, so even without a release, you can totally mix in your IDE/RCP the current release of Wild Web Developer with a newer release of LSP4E when it happens.

mickaelistria:
Actually, this change has no payload (except restricting decent compatibility with older LSP4E). I don't think we should restrain the ranges for bugfix, bugfixes come for free just by upgrading dependencies to their newer version in the installation, no need for all downstream consumers to take any action to allow that.

jcompagner:
yeah i am always a bit confused i kind of expected if it says:

org.eclipse.lsp4e;bundle-version="0.13.6"

that that is a "hard" version (because you have to play with [] or () to make ranges)
but it doesnt seem to be so, if you do that then OSGi also allows everything that is bigger?

i tried this in my installation and yes if i add this: https://download.eclipse.org/lsp4e/releases/latest/

and then upgrade the lib it works also, my problem is gone.

but the thing is:
https://download.eclipse.org/wildwebdeveloper/snapshots/plugins/
that already did contain the new lsp4e also as far as i can see
and i did update WWD, but that didn't help, i really had to do this manually

i guess because eclipse didn't find that as an upgrade? it did see the feature upgrade of WWD but then it still sees that it had the minimum requirements for WWD and didn't update lsp4e at the same time (i had todo that manually)

mickaelistria:
The bundle in download area is coming from the target-platform, which references the latest, so I guess just re-running the build would be sufficient for download area to include latest LSP4E. I've just retriggered a build.

mickaelistria:
Yes "0.13.6" is actually means "[0.13.6)", or "[0.13.6,0.0.0)" which means "at least 0.13.6".
Let's close as no change is needed.

jcompagner:
last night i think that https://download.eclipse.org/wildwebdeveloper/snapshots/plugins/ already had the latest lsp4e (0.13.8.xxx)

And i updated to that also already, but eclipse didn't get that 0.13.8 i still had an old one.
I guess because eclipse thought no need to update because its not needed.
That is a bit annoying for end users if you ask me.
Because how should they know that they are supposed to update just a plugin that is not really a feature directly but just something under the hood.

only when yesterday evening i really manually forced an update it worked. and it got the 0.13.8

mickaelistria:
Same issue can happen with any Eclipse plugin, it's not a specificity of Wild Web Developer. That's more an issue/limitation about p2's "Check for Updates" IMO. Currently, p2 does only look for newer versions of the "top-level" artifacts and doesn't look for newer version of dependencies.

jcompagner:
yes i know that it is not directly related to WWD 
but this makes issues that are reported quite horrible. because you don't really know what version a user is running on if they report i am running the latest WWD version X

it then can be any version of lsp4j or lsp4e underneath it

And i guess the current way to fix this is to fix the version number in WWD to LSP4E, but yes thats also not what we always want....

PR #792, opened 2022-02-17
Do not specify --tsserver-path

### LINKED ISSUE #651: Incompatible with Yarn 2 PnP linker
This is probably an issue with the TypeScript language server rather than WWD itself, but I'm not 100% sure.

I set up a project with Yarn 2, which uses the PnP linker by default, meaning that there is no node_modules directory. When I opened a TS file, I got errors about being unable to find modules that I'm importing from, even though the necessary dependencies are listed in package.json. I tried a few different things and finally "fixed" the issue by configuring Yarn to use the node-modules linker instead, so now there is a node_modules directory.

Weirdly, everything worked fine under the PnP linker for a .js file I was looking at. I'm not sure if the difference is .js vs .ts, or if it's because the .js used require() whereas .ts used import.

rdicroce:
Partially fixes #651.


mickaelistria:
Can you please also increase the version of org.eclipse.wildwebdeveloper by +0.0.1 in MANIFEST.MF and pom.xml as this is the 1st change since last release?

PR #803, opened 2022-03-21
Use newer DisplayHelper API

akurtakov:
And other JUnit 5-ification while at it.


mickaelistria:
One issue with ParameterizedTests is how they look on CI, it becomes almost impossible to figure out the actual issue from test reports. Can the display name or something be set?

akurtakov:
@mickaelistria I set the name but Tycho ignores it totally altough it changed in Eclipse .

mickaelistria:
I don't think it's something we can quickly fix in Tycho.
Can you please submit a PR without the ParameterizedTests so we merge those parts?

PR #810, opened 2022-04-03
Correct file extensions/names/patterns: ESLint, TypeScript config/lint

howlger:
In the [`org.eclipse.core.contenttype.contentTypes` extension point](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/extension-points/org_eclipse_core_contenttype_contentTypes.html), the `file-extensions` attribute must not contain `.` and `*` since _extension_ means here the string after the last `.` and the `file-names` attribute does not support the `*` wildcard in contrast to the `file-patterns` attribute.

Examples:
- `file-extensions="eslintrc"` matches the following file names: `.eslintrc`, `foo.eslintrc`, etc.
- `file-patterns="*.eslintrc.json"` matches the following file names: `.eslintrc.json`, `foo.eslintrc.json`, etc.
- `file-extensions="eslintrc.json"` does not match any filename

### DIFF
--- a/org.eclipse.wildwebdeveloper/plugin.xml
+++ b/org.eclipse.wildwebdeveloper/plugin.xml
@@ -57,8 +57,8 @@
       </content-type>
       <content-type
             base-type="org.eclipse.wildwebdeveloper.json"
-            file-extensions="*.eslintrc.json"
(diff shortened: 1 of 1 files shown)


mickaelistria:
Thanks!

PR #816, opened 2022-04-21
Update Vscode extensions

vrubezhny:
vscode                        1.63.2 --> 1.66.2

yaml-language-server           1.3.0  --> 1.7.0

### DIFF
--- a/org.eclipse.wildwebdeveloper/pom.xml
+++ b/org.eclipse.wildwebdeveloper/pom.xml
@@ -61,7 +61,7 @@
 							<goal>wget</goal>
 						</goals>
 						<configuration>
-							<url>https://az764295.vo.msecnd.net/stable/899d46d82c4c95423fb7e10e68eba52050e30ba3/code-stable-x64-1639562789.tar.gz</url>
+							<url>https://az764295.vo.msecnd.net/stable/dfd34e8260c270da74b5c2d86d61aee4b6d56977/code-stable-x64-1649664837.tar.gz</url>
 							<unpack>true</unpack>
 							<outputDirectory>${project.build.directory}/vscode</outputDirectory>
 						</configuration>
@@ -99,7 +99,7 @@
 							<goal>wget</goal>
 						</goals>
 						<configuration>
-                            <url>https://dbaeumer.gallery.vsassets.io/_apis/public/gallery/publisher/dbaeumer/extension/vscode-eslint/2.2.2/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage</url>
+                            <url>https://dbaeumer.gallery.vsassets.io/_apis/public/gallery/publisher/dbaeumer/extension/vscode-eslint/2.2.3/assetbyname/Microsoft.VisualStudio.Services.VSIXPackage</url>
 							<outputFileName>vscode-eslint-ls.zip</outputFileName>
 							<unpack>true</unpack>
 							<outputDirectory>${project.build.directory}/vscode-eslint-ls</outputDirectory>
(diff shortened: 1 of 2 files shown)


vrubezhny:
IP Team Review request is created for `typescript/4.6.3`:
 - https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2362
CQ Is created for `VSCode 1.66.2 (subparts)`:
 - https://dev.eclipse.org/ipzilla/show_bug.cgi?id=24057

For the other modules clearly defined reports:
`Vetted license information was found for all content. No further investigation is required.`

akurtakov:
Why using the old ipzilla? Shouldn't same requestbe made in the iplab ?

vrubezhny:
IP Team Review request is created for vscode/1.66.2:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2399

vrubezhny:
IP Team Review request is created for typescript/4.6.4:

https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2440

vrubezhny:
> Why using the old ipzilla? Shouldn't same requestbe made in the iplab ?

Opening a CQ here still may have sense as it's opened only for three LS extensions (provided in source code attached to the CQ) while the iplab request can be made only for the whole vscode source (at least currently  don't know how to make it for only the json/html/css extensions). Normally it takes more time to resolve a CQ than an iplab request, but in this certain case the license check is complete for the CQ (https://dev.eclipse.org/ipzilla/show_bug.cgi?id=24057) while the iplab request is still under investigation (https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/2399)

org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -93,9 +93,19 @@
...
+										<equals arg1="${github.api.token}" arg2="" />
+									</not>
+								</condition>
+								<then>
+									<headers>
+										<Authorization>token ${github.api.token}</Authorization>
+									</headers>
+								</then>
+							</if>
+
```
mickaelistria, line 108: How is it supposed to work? The wget mojo doesn't support `<if>` as far as I know, thos this whole block may be useless.

vrubezhny: when we don't provide a value for `${github.api.token}` (running it manually with `mvn clean verify`, f.i.,  or in WindowsPR build) this construction just doesn't add the `<headers>...</headers>` fragment to the effective pom.xml (so it's omitted).
When `${github.api.token}` is set to some value (like in Jenkins build) then `<headers>...</headers>` fragment is added to the effective pom.xml and that value is used in the requests header.

... or something like this.

mickaelistria: What makes you think this actually works? I'm not aware that `<if>` is a native Maven construct, nor that download-maven-plugin has support for this? The standard way to conditionally configure a pom.xml is to use Maven profiles.

vrubezhny: The example is taken from stackoverflow: https://stackoverflow.com/questions/10149539/conditional-configuration-in-maven-pom-xml
From the experience - we have two build configurations - WindowsPR and Jenkinsfile, also we can build locally running `mvn clean install`. So, after we added the github token as a parameter it become a required paramerter - as result local build with `mvn clean install` and WindowsPR build had started to fail (because we don't add this parameter to the commands. Jenkinsfile build started successfully download the required `package.json` file from `raw.githubuser...`. 
Adding the following `<if><condition>...<then>` construction:
```
							<if>
								<condition>
									<not>
										<equals arg1="${github.api.token}" arg2="" />
									</not>
								</condition>
								<then>
									<headers>
										<Authorization>token ${github.api.token}</Authorization>
									</headers>
								</then>
							</if>
```
made the local and WindowsPR builds to work normally.
So, I suppose it really works. 
If no `github.api.token` parameter value is provided, the `<headers>...</headers>` elements are not added to the effective pom.xml and, as result, wget action doesn't fail because of wrong header is sent (a default header is sent in this case)

mickaelistria:
> The example is taken from stackoverflow: https://stackoverflow.com/questions/10149539/conditional-configuration-in-maven-pom-xml

Please read the answers as well as the questions. This construct doesn't exist in Maven. Conditional configuration is achieved with profiles.

> Adding the following <if><condition>...<then> construction made the local and WindowsPR builds to work normally.

Because all that block is fully ignored, and the `Authorization` header is never added, and the download quotas are not exceeded yet for today to cause a download failure.

> So, I suppose it really works.

No, not at all :P

mickaelistria:
Can you please update this PR to include only LS update and create a separate PR for memory+docker settings?

mickaelistria:
@vrubezhny Can you please rebase? We may be able to merge those now.

PR #821, opened 2022-05-12
Fix github credencials (pass 2)

vrubezhny:
Github project credentials are to be used in `wget` action when accessing
`https://raw.githubusercontent.com` in order to minimize restriction limits


mickaelistria:
Log says `Masking supported pattern matches of $GITHUB_API_TOKEN`, this would need to be fixed. The `$...` expression may not be expanded in the `withCredentials` block, the direct value should be tried instead.

vrubezhny:
> Log says Masking supported pattern matches of $GITHUB_API_TOKEN, this would need to be fixed. The $... expression may not be expanded in the withCredentials block, the direct value should be tried instead.

Imho, it's OK and suits our needs - we don't want passwords/tokens to be visible in the log.

PR #825, opened 2022-05-19
Fix getting the ESLint LS package.json from GitHub

vrubezhny:
A workaround to avoid https://raw.githubusercontent.com download restriction limit


mickaelistria:
That seems good an working, so let's merge it to move forward.
However, I believe that to save some time, we could try the `export` goal instead of `checkout` as I believe the former skips history.

PR #834, opened 2022-06-14
Cross site scripting in Angular #2

vrubezhny:
Package @angular/core (npm) < 11.0.5

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestAngular.java
+++ b/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestAngular.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019 Red Hat Inc. and others.
+ * Copyright (c) 2019, 2022 Red Hat Inc. and others.
  *
  * This program and the accompanying materials are made
  * available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


vrubezhny:
At the moment it's OK that the `org.eclipse.wildwebdeveloper.tests.TestAngular.testAngular` JUnit test fails with the following stacktrace:
```
 `org.opentest4j.AssertionFailedError: Diagnostic not published in standalone component file ==> expected: <true> but was: <false>
	at org.eclipse.wildwebdeveloper.tests.TestAngular.testAngular(TestAngular.java:56)
```

The problem is that Angular LS in some circumstances doesn't report the Diagnostics for a project, which looks like a known issue. See: 
- https://github.com/angular/vscode-ng-language-service/pull/1668
- https://github.com/angular/vscode-ng-language-service/pull/1669

So, we probably need to wait these PRs to be released (probably in v. 14.1.0) and WWD to pick up the fixed Angular LS version.

mickaelistria:
IIRC, you identified some actions that make things work (something like opening editor, waiting for something, then closing, then reopening...). If it is right, can you please tweak the test to test that scenario and succeed more often.

vrubezhny:
The upstream issue is fixed in `@anhular/language-server/14.0.1, so the solution is to start using the fixed version of Angular LS. The v.14.0.1 is known to ClearlyDefined (`Vetted license information was found for all content. No further investigation is required.`) - so we're allowed to use it.

The test project vulnerabilities are fixed by using newer versions of the project dependencies

`TestAngular` JUnit test case is re-worked splitting it into a separate project setup and two testcases sharing the same project.
Using `old`version of  Angular LS was stably showing a failure for the first testcase, while the second one was succeeding. This behavior was due to the lack of diagnostics reporting by Angular LS when a project is just opened/loaded.

PR #837, opened 2022-06-23
NodeJSManager wrongly claims Node 16 is not supported

akurtakov:
And we shipt Node.js 16.x as our bundled.
The idea of static list of supported versions is plain wrong as we have
no idea which editor which version of Node.js supports it's best to
remove the static list and just say "recent" version is required when
using the "system" Node.


akurtakov:
Uncovered by Linux Tools build stuck at the Message dialog claiming Node.js 16 is not supported version.

Issue #851, opened 2022-08-22
language servers background jobs keep hanging around.


jcompagner:
Everytime i trigger a new project (so i open a ts file from a npm folder the first time)

i see a couple of jobs spinning up

![image](https://user-images.githubusercontent.com/3274268/185914675-b070146e-24f9-4dd6-a1a2-a06673899ac3.png)

that TS/JS thing will be quickly gone after a few seconds
but that other one keeps hanging around all the time, even after i close the last ts file again from that project..

it never goes away and it will add one everytime i reopen a ts file from a project where it then starts a language server for\

this is with the latest snapshot:

![image](https://user-images.githubusercontent.com/3274268/185915007-41ff6937-ddf4-4a38-9c58-58150600f62c.png)

jcompagner:
hmm could also be that this needs to go to LSP4e?

costescuandrei:
It happens to me as well.

jcompagner:
i will report this at LSP4e because i think this:

https://github.com/eclipse/lsp4e/blob/2632a5543974da44ec5ee942823ddf40c67bf104/org.eclipse.lsp4e/src/org/eclipse/lsp4e/progress/LSPProgressManager.java#L99

will never be hit for that job. so there is never an "End"

jcompagner:
closing this one for now because it seems to be a lsp4e problem

jcompagner:
[test.zip](https://github.com/eclipse/wildwebdeveloper/files/9451607/test.zip)

brand new workspace dir
extract there that zip.
do a npm install once to create the node_modules
then import the "test" project into eclipse
open also the progress view (so you see it better)

open src/app/app.component.ts

then you see right away a few jobs one of them being "Language Server Background Job"
which is also really constantly in the status bar progress area.

now close the app.component.ts

the "Language Server Background Job" is not going away (and its also still in the status bar)

open again app.component.ts 

and you get another one...

and it is not that something else hangs like the node js or the console windows

![image](https://user-images.githubusercontent.com/3274268/187402724-bb916e59-d767-44c3-9e7a-d605dcf42791.png)


those i have when i have opened 1 ts file 
and if close the ts file again (so i don't have anything open for that project) all the sub processes are all cleaned up
so that works nicely..

mickaelistria:
I can reproduce the issue and am investigating.

jcompagner:
reopening because the fix needs to be in the WWD AngularClient impl:

[...]

PR #855, opened 2022-08-31
fix for language servers background jobs keep hanging around. #851

jcompagner:
remove the now implemented notifyProgress in the super class


mickaelistria:
Thanks!

PR #858, opened 2022-09-05
XML: Closing tags should be included in the code folding range

### LINKED ISSUE #857: XML: Closing tags should be included in the code folding range
It would be nice if in the **XML** editor the **end tag** of a folded element would be **included in the folding** (which is currently not the case in Eclipse in the _Generic Text Editor_), in the same was as described in [LemMinX issue #1178](https://github.com/eclipse/lemminx/issues/1178) (which has already been implemented and seems to work in VSCode).

angelozerr:
XML: Closing tags should be included in the code folding range

Fixes #857


mickaelistria (approved):

mickaelistria:
Thank you!

PR #861, opened 2022-09-06
Support for xml.downloadExternalResources.enabled

### LINKED ISSUE #781: Support for xml.downloadExternalResources.enabled
We have added support for `xml.downloadExternalResources.enabled` which controls whether the language server will download an external resource that is requested as a result of loading some document. For example, an external DTD is required when opening an XML file.

See https://github.com/eclipse/lemminx/pull/1155 .

angelozerr:
Support for xml.downloadExternalResources.enabled

Fixes #781


angelozerr:
This PR add the external download resources settings which is very important for security. By default it is enabled (the download of XSD , DTD will occur like today,so it willchange nothing). But if you disable it, download is forbidden and you have a code action will give the capability to download it). See the demo below:

DownloadExternaResourcesDemo

PR #862, opened 2022-09-07
OnEnterRules support for XML

angelozerr:
OnEnterRules support for XML

### DIFF
--- a/org.eclipse.wildwebdeveloper.xml/language-configurations/xml/xml.language-configuration.json
+++ b/org.eclipse.wildwebdeveloper.xml/language-configurations/xml/xml.language-configuration.json
@@ -14,17 +14,26 @@
 		["<", ">"],
 		["'", "'"],
 		["\"", "\""]
+	],
+	"onEnterRules": [
+		{
+			"beforeText": {
+				"pattern": "<([_:\\w][_:\\w-\\.\\d]*)(?:(?:[^'\"/>]|\"[^\"]*\"|'[^']*')*?(?!\\/)>)[^<]*$",
+				"flags": "i"
+			},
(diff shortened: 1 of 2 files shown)


angelozerr:
This PR requires https://github.com/eclipse/tm4e/pull/438 

It declare on enter values for XML / XSL and improve the editing of an XML file by auto indenting and create a new line when user type on Enter.

Here a demo without on enter rules (current master):

WithoutOnEnterRule

Here a demo with on enter rules (the PR):

WithOnEnterRule

mickaelistria:
> This PR requires https://github.com/eclipse/tm4e/pull/438

We shouldn't depend on this until we release TM4E. So the question is whether having this commit without having  https://github.com/eclipse/tm4e/pull/438 does cause some annoying bug?
If it does not cause any bug, then we can merge now so it would partially work (or even not work) but won't cause issue, and this will only fully work with newer TM4E.

angelozerr:
>We shouldn't depend on this until we release TM4E. So the question is whether having this commit without having https://github.com/eclipse/tm4e/pull/438 does cause some annoying bug?

Yes can merge it, because onEnterRule is ignored if there is no indent declaration. Before my PR `indentAction` was used instead of `indent` and if you see the code https://github.com/eclipse/tm4e/pull/438/files#diff-8ea1b94a04d7dfa972f1916f2f1b2576214e37d9fe49a542c7d8bc9f0a8cd5a8R67 it doesn't create onEnterRule if there is no indent information.

> If it does not cause any bug, then we can merge now so it would partially work (or even not work) but won't cause issue, and this will only fully work with newer TM4E.

It will not cause some bug and it will not work.

PR #864, opened 2022-09-08
OnEnterRules support for html files

angelozerr:
OnEnterRules support for html files


mickaelistria:
I think we need a CQ to allow inclusion of this language-configuration file. I created https://dev.eclipse.org/ipzilla/show_bug.cgi?id=24255

angelozerr:
Why we need not a CQ for JS/TS OnEnterRules file?

mickaelistria:
> Why we need not a CQ for JS/TS OnEnterRules file?

We already had a CQ and approval for those when we included the language-configuration for the 1st time. The changes are then increments without strong IP matter, so no CQ required.

angelozerr:
The languag econfiguration for HTML already exists at https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/language-configurations/html/language-configuration.json

I have just renamed to `language-configuration.jsonc`  (like I did for JST/TS) to avoid error with comments when you open it in Eclipse IDE.

mickaelistria:
Ah OK, my bad. So no need for a CQ indeed!

PR #903, opened 2022-10-10
Use tycho 3.1.0

vrubezhny:

### DIFF
--- a/org.eclipse.wildwebdeveloper.embedder.node.macos.aarch64/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper.embedder.node.macos.aarch64/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: Node.js for MacOS AArch64
 Bundle-SymbolicName: org.eclipse.wildwebdeveloper.embedder.node.macos.aarch64
-Bundle-Version: 0.1.3.qualifier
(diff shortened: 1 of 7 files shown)


akurtakov:
Not that I am against this change but why?

vrubezhny:
> Not that I am against this change but why?

It looks like tycho v.3.0.0-SNAPSHOT was removed from its repository.
Some tycho plugins have changed their location (moved to parent from `extra/`) so it appeared having v.3.0.0 versions for those plugins in their old locations, while v.3.1.0 versions appeared in new location
It looks like these changes as well as some changes in tycho v.3.1.0 caused some other build failures to appear on WWD PR builds, like this one:

```
[ERROR] Unresolveable build extension: Plugin org.eclipse.tycho:tycho-maven-plugin:3.0.0-SNAPSHOT or one of its dependencies could not be resolved: Could not find artifact org.eclipse.tycho:tycho-maven-plugin:jar:3.0.0-SNAPSHOT in cbi (https://repo.eclipse.org/content/groups/cbi) @ 
```

vrubezhny:
Reasonable, but then we'll be switching it again to use snapshot tycho build?
Also, tycho releases has the same problem - `tycho-buildtimestamp-jgit` plugin has changed its placement - so any update to use tycho v.3.1.0 will require the same changes as I did now.

I can rollback this change and switch WWD to use tyycho release v.3.0.0 instead of snapshot, but, honestly, I see no reason to to this. Switching to release v.3.1.0 is possible but still requires the same change to be merged.

PR #913, opened 2022-10-21
HTML preferences page

angelozerr:
HTML preferences page

### DIFF
--- a/org.eclipse.wildwebdeveloper/plugin.properties
+++ b/org.eclipse.wildwebdeveloper/plugin.properties
@@ -13,3 +13,7 @@ pluginName=Wild Web Developer: web development in Eclipse IDE
 providerName=Eclipse Wild Web Developer project
 
 HTMLPreferencePage.name=HTML (Wild Web Developer)
+HTMLCompletionPreferencePage.name=Completion
+HTMLFormatPreferencePage.name=Format
+HTMLHoverPreferencePage.name=Hover
+HTMLValidationPreferencePage.name=Validation
--- a/org.eclipse.wildwebdeveloper/plugin.xml
+++ b/org.eclipse.wildwebdeveloper/plugin.xml
@@ -283,6 +283,7 @@
       <server
             class="org.eclipse.wildwebdeveloper.html.HTMLLanguageServer"
             serverInterface="org.eclipse.wildwebdeveloper.html.HTMLLanguageServerAPI"
+            clientImpl="org.eclipse.wildwebdeveloper.html.HTMLLanguageClient"
             id="org.eclipse.wildwebdeveloper.html"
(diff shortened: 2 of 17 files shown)


angelozerr:
This PR provides almost all settings of HTML langugae server for format, completion, hover and validation:

image

I know that label is too large, but I generate that from package.nls.json of vscode html. Don't hesitate to review those labels.

I need to comment my code.

The interesting code is https://github.com/eclipse/wildwebdeveloper/pull/913/files which should be generic for all language server which track preference store and call didChangeConfiguration.

The main idea is that I will extend this class for CSS and naother language server (ex : XML too)

The another class which is important is https://github.com/eclipse/wildwebdeveloper/pull/913/files#diff-a7f8d0a980c6f4c5d393048aef925f7f69120eb5cc8fcbb624193fcb006e9375 which overrides configuration to return the proper HTML configuration.

I need to add comments with vscode html link code to explain why we need to do that.

Here a litlle demo with validation preference page:

HTMLPreferences

Hope you will like it.

angelozerr:
@mickaelistria the PR is ready to review it. Is there any chance that you review it? I'm waiting for merge of this PR to do the same thing for CSS by providing all CSS preferences supported by the CSS language server.

akurtakov:
@vrubezhny Please review as @mickaelistria is not available this week.

vrubezhny:
> I know that label is too large, but I generate that from package.nls.json of vscode html. Don't hesitate to review those labels.

@angelozerr Why not to edit these long text labels by splitting them into several lines just using the new line char to decrease the width of the preference page? For example, like this:

Screenshot from 2022-10-26 19-06-59

INMHO ,this would give some extra space to the value editors.
And later we can think of adding a link to open the provided link in a browser and making the label to be shorter. and more useful, because, as the label text is not possible to be selected and copied to the clipboard, a user will need to manually re-type it in to a browser to see the content,

angelozerr:
> @angelozerr Why not to edit these long text labels by splitting them into several lines just using the new line char to decrease the width of the preference page? For example, like this:

Thanks a lot for your suggestion, fixed.

angelozerr:
> INMHO ,this would give some extra space to the value editors.
And later we can think of adding a link to open the provided link in a browser and making the label to be shorter. and more useful, because, as the label text is not possible to be selected and copied to the clipboard, a user will need to manually re-type it in to a browser to see the content,

I agree UI preferences should be improved, but my main goal is to provide teh capability to customize HTML language server with Eclipse preferences. The UI could be improved later.

vrubezhny (approved):
Looks good. Thanks!

angelozerr:
Thanks so much @vrubezhny !

Issue #932, opened 2022-10-31
[a11y] Accessibility on links in JAWS doesn't always work


GeraldMit:
JAWS for screen reading in Eclipse is sometimes not reading Links.

Adding an AccessibleAdapter and explicitly setting the readable text seems to fix the issue.

Ex. In the XML Preference page

For
		`catalogsLink.setText(Messages.XMLPreferencePage_XMLCatalogsLink);`
then add
```
                catalogsLink.getAccessible().addAccessibleListener(new AccessibleAdapter() {
			@Override
			public void getDescription(AccessibleEvent event) {
				event.result = catalogsLink.getText();
			}
		});
```

I tracked the links in WildWebDeveloper and added the AccessibleAdapter, and I am testing.

mickaelistria:
> JAWS for screen reading in Eclipse is sometimes not reading Links.

Isn't it more a Platform/SWT issue then?

GeraldMit:
I was narrow-scoping, thanks @mickaelistria I should be looking at Link

PR #935, opened 2022-10-31
Issue #932

GeraldMit:
Changes to ChromeExecutableTab, FirefoxAttachTab, XMLPreferencesPage, and SchemaAssociationsPreferencePage to add Accessibility support for some versions of JAWS screen reader to resolve Issue #932


mickaelistria:
I'm unsure here; isn't `getText` supposed to be the default used for accessibility? I would expect those wouldn't be necessary to repeat and that SWT would behave that way by default?
How do other debugger tabs manage it?

GeraldMit:
> I'm unsure here; isn't `getText` supposed to be the default used for accessibility? I would expect those wouldn't be necessary to repeat and that SWT would behave that way by default? How do other debugger tabs manage it?

I didn't think to test them: I was focused on WWD and I was narrow-scoping the problem resolution; I will investigate in SWT at a higher level.

Thank you @mickaelistria

PR #937, opened 2022-11-02
Preference store per language

angelozerr:
Preference store per language

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/Activator.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/Activator.java
@@ -12,10 +12,16 @@
  *******************************************************************************/
 package org.eclipse.wildwebdeveloper;
 
+import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.PlatformUI;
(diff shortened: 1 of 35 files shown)


angelozerr:
This PR provides the capability to have a preference store per language ID.

This PR is very important (please merge it before the release) to support in the future spacesForTab preferences per language.

angelozerr:
This PR doesn't work -( since it seems that we can have a preference store per plugin. It is really annoying.

mickaelistria:
> This PR doesn't work -( since it seems that we can have a preference store per plugin. It is really annoying.

Do you mean that the current code ignores the store, fails at creating it, or throws other errors...?
If this helps, you may consider creating another scope dedicated to Wild Web Developer prefs.

mickaelistria:
One possibility would be to create some kind of SubPreferenceStore that would allow to use a single store, but have a projection and subset of it. So when asking for preference `tabsWidth`, such SubPreferenceStore would actually request `languageServerId/tabWith` on the preferenceStore of org.eclipse.wildwebdeveloper.

angelozerr:
@mickaelistria we need to use `spacesForTab` and not `languageServerId/tabWith` as key, because this settings should be used:

 * by LSP4E for formatting
 * by the editor when user will press on `Tab`, it should generate some spaces (if spacesForTab is enabled)
 * by TM4E with onEnterRules

For the 2 last usecase, `languageServerId/tabWith` is unkwown.

I can manage those all usecases with https://github.com/eclipse-platform/eclipse.platform.text/pull/109

akurtakov:
What is the status of this one? PRs lurking around are generally not helpful if no one plans to work on them.

angelozerr:
Sorry I have no time to work on it.

PR #967, opened 2022-11-17
Npm license check

mickaelistria:


vrubezhny:
We have to add some exclusions to the check, otherwise we'll have this check failing all the time (I mean vscode extensions, eslint ls etc - those we download manually)

mickaelistria:
What's the status of das-license tool nowadays regarding local npm modules in package-lock.json? Should ebasing just work?

vrubezhny:
> What's the status of das-license tool nowadays regarding local npm modules in package-lock.json? Should ebasing just work?

In order to make it successfully verify it's probably needed to add ` -excludeSources local` parameter - then it will not be failing on local sources (it'll just skip them)

Next step would be adding the support for `/request-license-review'`-like comment which should run the dash-license tool with parameters forcing the review creation  ` -review -project tools.wildwebdeveloper -token ${{ secrets.GITLAB_API_TOKEN }}`

vrubezhny:
When I run it from `~/.m2/repository` (exactly as the workflow does) I get the following error:

```
 $ java -jar $(ls ~/.m2/repository/org/eclipse/dash/org.eclipse.dash.licenses/0.0.1-SNAPSHOT/org.eclipse.dash.licenses-*.jar | head -n 1) ./package-lock.json -excludeSources local -summary target/dash/npm-deps-summary
Can't write to target/dash/npm-deps-summary
```
and it doesn't depend on whether `./target/dash/npm-deps-summary` directory exists or not.
 
When I omit the `-summary` argument I get it working almost as expected:
```
$ java -jar $(ls ~/.m2/repository/org/eclipse/dash/org.eclipse.dash.licenses/0.0.1-SNAPSHOT/org.eclipse.dash.licenses-*.jar | head -n 1) ./package-lock.json -excludeSources local 
[main] INFO Querying Eclipse Foundation for license data for 185 items.
[main] INFO Found 51 items.
[main] INFO Querying ClearlyDefined for license data for 134 items.
[main] INFO Found 134 items.
[main] INFO Vetted license information was found for all content. No further investigation is required.
```

mickaelistria:
And it now works!

vrubezhny:
I added https://github.com/eclipse/dash-licenses/issues/196 in order to track this `-summary` argument related error, however I don't think we really need that argument to be used for us

PR #977, opened 2022-11-23
Use GitLab Token for License Check

vrubezhny:


.github/workflows/licensecheck.yml
```diff
@@ -34,4 +34,4 @@ jobs:
           ${{ runner.os }}-maven-
     - name: License check
       run: | 
-        mvn -U -V -e -B -ntp org.eclipse.dash:license-tool-plugin:license-check --file pom.xml
+        mvn -U -V -e -B -ntp org.eclipse.dash:license-tool-plugin:license-check --file pom.xml -Ddash.projectId=tools.wildwebdeveloper -Ddash.iplab.token=${{ secrets.GITLAB_API_TOKEN }}
```
mickaelistria, line 37: Maybe it's worth also using this change to enable the common license check action, as done in Tycho and m2e: https://github.com/eclipse-tycho/tycho/blob/master/.github/workflows/licensecheck.yml

vrubezhny: Sure.
It's just that I'm not sure if the token really works... The `dash.iplab.token` argument was accepted for sure because the Dash Licence plugin reported on that IPTeam issues are already exist:
```
[INFO] A review request already exists https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/5062 .
...
[INFO] A review request already exists https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/5063 .
```
But it never tried to create an issue yet, so I don't know yet if the argument value is really accepted.

vrubezhny: What I'm not sure about is secret name I've used... The guide and examples (like https://github.com/eclipse-tycho/tycho/blob/master/.github/workflows/licensecheck.yml) stay on adding a project name to the token (like `TYCHO_GITLAB_API_TOKEN`, for example), while Frederic states he created `GITLAB_API_TOKEN` (NOT prefixed with the WWD project name).  Probably I'm missing something.

vrubezhny: Also, we probably have to add the token usage to NPM dependencies check in https://github.com/eclipse/wildwebdeveloper/pull/967

akurtakov:
Your command is not hooked to anything in licensecheck.yml thus nothign will happen for /request-license-review . It is handled at https://github.com/eclipse/dash-licenses/blob/master/.github/actions/maven-license-check-action/action.yml#L29 so using it will have effect.

vrubezhny:
> Your command is not hooked to anything in licensecheck.yml thus nothign will happen for /request-license-review . It is handled at https://github.com/eclipse/dash-licenses/blob/master/.github/actions/maven-license-check-action/action.yml#L29 so using it will have effect.

@akurtakov  aren't we using it  when we invoke the following job at https://github.com/eclipse/wildwebdeveloper/pull/977/files#diff-214dc48999f71aa20bbf3110511b812571737db7a5da2e71a94d16c55255c08fR40: 

```
 call-license-check:
    uses: eclipse/dash-licenses/.github/workflows/mavenLicenseCheck.yml@master
    ...
```
?

mickaelistria:
> so, it looks like we're double checking them

The idea is to replace the custom `mvn` step by invoking the GitHub workflow. As to `/request-license-review`, it will most likely only work when the necessary config is merged to master.

vrubezhny:
> > so, it looks like we're double checking them
> 
> The idea is to replace the custom `mvn` step by invoking the GitHub workflow. As to `/request-license-review`, it will most likely only work when the necessary config is merged to master.

So, we should remove the 'build` job as result, right?

How can I make sure that the dependency list generated in the `build` job equals to the one that is generated in "the GitHub workflow"? I mean, currently they produce quite the same list of not vetted artifacts, but probably the best way would be comparing the generated lists of dependencies to be used by License Check?

mickaelistria:
> So, we should remove the 'build` job as result, right?

Yes, the 'build' job would "only" build and test; and reviewing the license would be another workflow (like it's done in Tycho and m2e). This becomes easier to maintain and review.

> How can I make sure that the dependency list generated in the build job equals to the one that is generated in "the GitHub workflow"?

You don't have to make sure of that, but only to trust the GitHub workflow (Which is becoming the standard approach recommended by CBI).

vrubezhny:
cc @wildwebdeveloper-bot

mickaelistria:
Again "/request-license-review" is not likely to work before the change is merged. A 1st iteration is to make the GitHub workflow use the CBI one; then we'll see if /request-license-review can work with it.

vrubezhny:
@mickaelistria  OK. Then I'm going to merge the change that replaces the `build` job with the ` call-license-check` that invokes the dash-licenses's `mavenLicenseCheck.yml` and enables the invocation on `issue-comment` event

.github/workflows/licensecheck.yml
```diff
@@ -12,26 +12,34 @@ on:
     branches: 
      - 'master'
      - 'tycho-*'
```
mickaelistria, line 14: You can probably remove it by the way.

.github/workflows/licensecheck.yml
```diff
@@ -12,26 +12,34 @@ on:
...
+#      uses: actions/cache@v3
+#      with:
+#        path: ~/.m2/repository
+#        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+#        restore-keys: |
+#          ${{ runner.os }}-maven-
+#    - name: License check
+#      run: | 
+#        mvn -U -V -e -B -ntp org.eclipse.dash:license-tool-plugin:license-check --file pom.xml -Ddash.projectId=tools.wildwebdeveloper -Ddash.fail=true  -Ddash.iplab.token=${{ secrets.GITLAB_API_TOKEN }}
+#
```
mickaelistria, line 39: just remove it. If we need it back, we can revert.

.github/workflows/licensecheck.yml
vrubezhny: OK

PR #978, opened 2022-11-24
Add Maven Coordinates for Jakarta dependency artifacts

vrubezhny:


mickaelistria:
Thanks! let's see if those do take higher priority and then make the license check happy by generating proper coordinates.

mickaelistria:
And yes, it worked!

PR #987, opened 2022-12-04
Allow tsconfig.json to have trailing commas

### LINKED ISSUE #971: Feature request: Allow `tsconfig.json` to have trailing commas
WWD cannot start a debug session for a `*.ts` file unless its corresponding `tsconfig.json` is free of trailing commas & comments (i.e. it must be "true" JSON).  In contrast, `tsc` and the TypeScript language server (or at least recent versions of them) have no trouble with dealing with that.  It must be they allow the `tsconfig.json` to use the syntax of a `*.jsonc` / "JSON with comments" file.

This means it may be possible to build & run a TypeScript project in a terminal, but not debug it in WWD (unless attaching to a running instance), because WWD's requirements on what constitutes a valid `tsconfig.json` are stricter than `tsc`'s.

To save the effort of having to edit a `tsconfig.json` just to make it compatible with WWD, it would be nice if WWD could be more lax in its parsing of the `tsconfig.json` file, to parse only the bare minimum of what it needs to start a debug session.

AndrewFerr:
Fixes #971

---

Note that I did look into replacing Gson with another parser (I gave [Jackson](https://github.com/FasterXML/jackson) a try), but it turns out that Gson is used not only in WWD, but across the LSP4E and TM4E projects, which all pass Gson objects between each other.  Replacing Gson would thus take a bit of coordination across all three projects.

Plus, the libraries of whatever new parser is used would have to be provided in WWD's Eclipse repository, as Gson currently is.

So for now, just doing some manual parsing should be the least disruptive solution for the filed issue.

Also note that the comment-stripping isn't strictly necessary (as Gson's default configuration accepts comments), but it helps with handling comments that follow a trailing comma.

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/node/NodeRunDAPDebugDelegate.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/node/NodeRunDAPDebugDelegate.java
@@ -26,6 +26,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.regex.Pattern;
 
 import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.resources.IFile;
@@ -305,6 +306,9 @@ public class NodeRunDAPDebugDelegate extends DSPLaunchDelegate {
 		return null;
 	}
 	
+	private static final Pattern BlockCommentPattern = Pattern.compile("(?<!//.*)/\\*(?:.|\\R)*?\\*/");
+	private static final Pattern LineCommentPattern = Pattern.compile("\\s*//.*");
+	private static final Pattern TrailingCommaPattern = Pattern.compile(",(\\s*)\\}");
 	public Map<String, Object> readJSonFile(File tsConfgFile) {
 		if (tsConfgFile == null || !tsConfgFile.isFile()) {
 			return Map.of();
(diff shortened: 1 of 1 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/node/NodeRunDAPDebugDelegate.java
```diff
@@ -316,7 +320,11 @@ public Map<String, Object> readJSonFile(File tsConfgFile) {
 				response.append(inputLine).append('\n');
 			}
 			Type type = new TypeToken<Map<String, Object>>() {}.getType();
-			return new Gson().fromJson(response.toString(), type);
+			var responseString = response.toString();
+			responseString = BlockCommentPattern.matcher(responseString).replaceAll("");
+			responseString = LineCommentPattern.matcher(responseString).replaceAll("");
+			responseString = TrailingCommaPattern.matcher(responseString).replaceAll("$1}");
```
mickaelistria, line 326: Can you please extract it into a dedicated method with a nice name and some documentation to not forget what case they're serving?

mickaelistria (changes requested):

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/node/NodeRunDAPDebugDelegate.java
AndrewFerr: Done by acdd258.  The docstring doesn't make explicit mention of what kind of preprocessing is done, so that if any other JSON tweaks are needed, the documentation & method name may remain unchanged & continue to be accurate.

mickaelistria:
Thanks. It looks like there is some ECA check issue. Can you please verify you used on both commits the author email that matches the one you signed ECA with?

AndrewFerr:
I've amended the commits' author/committer email address to the one I signed the ECA with.

mickaelistria:
The failure is definitely not related. let's merge. Thanks!

PR #998, opened 2022-12-12
Set dependabot Open PR limit for NPM Module updates

vrubezhny:


mickaelistria:
What's the benefit with this change?

vrubezhny:
It's just that the last update (for, typescript-language-server@2.2.0` was published ~2 days ago, but dependabot here didn't created a PR for it until this evening. So, I thought about this limit.

IMHO, it's still good to set it up to 10, but it's not relevant to the late update of `typescript-language-server`

vrubezhny:
Note that by default `open-pull-requests-limit` is 5. So, as far as I understand, having 5 opened dependabot update PRs in the queue will prevent the creation of PR for the new pending updates.

mickaelistria (approved):
OK, thanks. Feel free to merge then,

PR #992, opened 2022-12-08
Use commons-compress from upstream

mickaelistria:

PR #1004, opened 2022-12-14
Switch to commons-compress from Maven.

akurtakov:
To get licensecheck clean.


akurtakov:
Looks like Mickael almost did it already https://github.com/eclipse/wildwebdeveloper/pull/992

PR #1009, opened 2022-12-15
New XML format settings

angelozerr:
New XML format settings

Here a screenshot will all missing settings coming from the new formatter:

image

Label should be revisisted because some label are too long (any review are welcome), but I did that quickly .

### DIFF
--- a/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/ui/Messages.java
+++ b/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/ui/Messages.java
@@ -31,6 +31,8 @@ public class Messages extends NLS {
 	public static String XMLCodelensPreferencePage_codelens_enabled;
 
 	// --------- XML Formatting preference page
+	public static String XMLFormattingPreferencePage_format_maxLineWidth;
+	public static String XMLFormattingPreferencePage_format_grammarAwareFormatting;
 	public static String XMLFormattingPreferencePage_format_emptyElements;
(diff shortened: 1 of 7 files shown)


garretwilson:
> Label should be revisisted because some label are too long (any review are welcome), but I did that quickly .

Just a tip; things like "The value represents the …" should not be part of the label. That is help text. A long time ago we would have mouse hovers over the label or over a question mark that would show the help text as a tooltip. I note also there is a (?) in the bottom, left-hand corner to bring up help, but I don't know if anybody really uses the help system nowadays.

At the least you should probably move this help text to a separate label. But most of these could use extra explanation, so you can't add extra labels for all of them.

Surely Eclipse in general has provided some policy for how plugins should provide help information in the preferences dialog? I would be willing to bet on it, as surely this is not the first time a plugin ever needed to provide help information for an option.

garretwilson:
> … I have no time to do that …

I completely understand. But then I suggest taking out the "help sentence" altogether. Users might find it irritating when they change to another entry in the preferences dialog and suddenly the dialog width doubles or triples (or shows scroll bars, which would be even more irritating).

garretwilson:
"Join CDATAContent on format" has a typo: a missing space.

mickaelistria:
Let's merge it as it's good quality.
@garretwilson your suggestions for labels are more than welcome as further pull requests!

angelozerr:
> "Join CDATAContent on format" has a typo: a missing space.

Please contribute by update any label at https://github.com/eclipse/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/ui/messages.properties#L28

PR #1058, opened 2023-01-16
Erroneous catalog

mickaelistria:


mickaelistria:
@CarlosMOGoncalves this should be available in next snapshots within a few minutes. After that, please install latest snapshots when available and report whether things are still not working.

CarlosMOGoncalves:
Hello @mickaelistria ,

I have just installed the snapshot and indeed it fixes the issues I reported.
In short, whatever is missing from the Payara plugin is no longer preventing the XML editor from performing normally, navigation included.

Thanks a lot for your work and also for pointing it out on Payara repo.

PR #1067, opened 2023-01-24
'Unknown at rule' on file extensions associated with content type LESS

### LINKED ISSUE #1017: 'Unknown at rule' on file extensions associated with content type LESS
The framework I use uses the `.variables` extension to declare less variables. I went to the menu `Window -> Preferences -> General -> Content Types -> Text -> Wild Web Developer supported web dev file -> LESS` and added the `*.variables` extension.

When I create a `test.variables` file and add the content:

```less
@horizontalPadding: 1.0em;
```

It complained:

`Unknown at rule @horizontalPadding`

This error is not shown on .less files.

angelozerr:
'Unknown at rule' on file extensions associated with content type LESS

Fixes #1017

### DIFF
--- a/org.eclipse.wildwebdeveloper/plugin.xml
+++ b/org.eclipse.wildwebdeveloper/plugin.xml
@@ -198,15 +198,18 @@
       </server>
       <contentTypeMapping
             contentType="org.eclipse.wildwebdeveloper.css"
-            id="org.eclipse.wildwebdeveloper.css">
+            id="org.eclipse.wildwebdeveloper.css"
+            languageId="css">
       </contentTypeMapping>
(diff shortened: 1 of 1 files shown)


angelozerr:
The CSS language server uses languageId to know if the file is a CSS, LESS, or SCSS 

See https://github.com/microsoft/vscode/blob/06bd7ae776e2706a4aba2690d95788327dd16c43/extensions/css-language-features/server/src/cssServer.ts#L140

S we need to force the languageId in plugin.xml in order to give the capability for new content type which extends to LESS / SCSS to apply correctly LESS / SCSS rules

mickaelistria:
Thanks

Issue #1070, opened 2023-01-26
Typescript analysis broken after update


ZeeD:
After updating wwd I have this popup in the lower left corner with the error:

```
LSP (JavaScript-TypeScript Language Server)
Inlay Hints request failed. Requires TypeScript 4.4+.
```

every time I open a `.ts` file. and then the autocomplete is broken (the syntax colors still works)

a quick check on the running processes showed me

```
$ ps aux | grep typescript
a206719281       29496   0,0  1,2  7189320 392144 s000  T    10:02am   0:06.18 /Users/a206719281/eclipse/java-latest/Eclipse.app/Contents/Eclipse/.node/node-v18.12.1-darwin-x64/bin/node /Users/a206719281/eclipse/java-latest/Eclipse.app/Contents/Eclipse/../../../../../.p2/pool/plugins/org.eclipse.wildwebdeveloper_1.0.1.202211171559/node_modules/@angular/language-server/index.js --ngProbeLocations /Users/a206719281/eclipse/java-latest/Eclipse.app/Contents/Eclipse/../../../../../.p2/pool/plugins/org.eclipse.wildwebdeveloper_1.0.1.202211171559/node_modules/@angular/language-service --tsProbeLocations /Users/a206719281/eclipse/java-latest/Eclipse.app/Contents/Eclipse/../../../../../.p2/pool/plugins/org.eclipse.wildwebdeveloper_1.0.1.202211171559/node_modules/typescript --stdio --logVerbosity terse
a206719281       29605   0,0  0,0  4259848    420 s000  R+   10:03am   0:00.00 grep --color=auto typescript
a206719281       29502   0,0  0,2  6006784  72476 s000  T    10:02am   0:00.33 /Users/a206719281/eclipse/java-latest/Eclipse.app/Contents/Eclipse/.node/node-v18.12.1-darwin-x64/bin/node /Users/a206719281/eclipse-workspace/ocellus-js-sdk-src/node_modules/typescript/lib/typingsInstaller.js --globalTypingsCacheLocation /Users/a206719281/Library/Caches/typescript/3.9 --typesMapLocation /Users/a206719281/eclipse-workspace/ocellus-js-sdk-src/node_modules/typescript/lib/typesMap.json --validateDefaultNpmLocation
[...]

PR #1071, opened 2023-01-26
Typescript analysis broken after update #1070

vrubezhny:
The TypeScript Language Server is set to use the WWD provided 'tsserver.js'

Issue: #1070

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/JSTSLanguageServer.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/JSTSLanguageServer.java
@@ -42,6 +42,8 @@ public class JSTSLanguageServer extends ProcessStreamConnectionProviderWithPrefe
 
 	private static final String[] SUPPORTED_SECTIONS = { "typescript", "javascript" };
 
+	private static String tsserverPath;
+	
 	public JSTSLanguageServer() {
 		super(JSTS_LANGUAGE_SERVER_ID, Activator.getDefault().getPreferenceStore(), SUPPORTED_SECTIONS);
 		List<String> commands = new ArrayList<>();
@@ -49,10 +51,12 @@ public class JSTSLanguageServer extends ProcessStreamConnectionProviderWithPrefe
 		try {
 			URL url = FileLocator
 					.toFileURL(getClass().getResource("/node_modules/typescript-language-server/lib/cli.mjs"));
+			File nodeModules = new File(url.getPath()).getParentFile().getParentFile().getParentFile();
+			tsserverPath = new File(nodeModules, "typescript/lib/tssserver.js").getAbsolutePath();
 			commands.add(new File(url.getPath()).getAbsolutePath());
 			commands.add("--stdio");
-			URL nodeDependencies = FileLocator.toFileURL(getClass().getResource("/"));
 			setCommands(commands);
+			URL nodeDependencies = FileLocator.toFileURL(getClass().getResource("/"));
 			setWorkingDirectory(nodeDependencies.getPath()); // Required for typescript-eslint-language-service to find
 																// it's dependencies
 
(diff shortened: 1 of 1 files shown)


mickaelistria:
As mentioned on the bug, I'm worried about what happens if a project uses a newer TypeScript and Wild Web Developer isn't updated. Would the newer typescript/tsserver be used? That is the expectation.

vrubezhny:
TypeScript Language Server used by WWD's JS/TS Editor should use a compatible version of `tsserver.js`  - the compatible version for sure is provided with the WWD installation - it's tested and more or less fresh. Anyway the JS TS Editor should use use exactly the version that is bundled with WWD.
User projects may use any `tsserver.js`  version, when they run/executed - JS TS Editor's  TypeScript LS is not involved here, IMHO.

mickaelistria:
> TypeScript Language Server used by WWD's JS/TS Editor should use a compatible version of tsserver.js - the compatible version for sure is provided with the WWD installation - it's tested and more or less fresh. Anyway the JS TS Editor should use use exactly the version that is bundled with WWD.

I disagree. Using a newer tsserver can provide a huge feature gain to developers without having them forced to update Wild Web Developer.
By choosing this strategy, you basically require more effort from people who move forward with versions than for people who stick to older versions. Given that the latest category are often doing a disservice to the ecosystem; I wouldn't favor those around the former.

> User projects may use any tsserver.js version, when they run/executed - JS TS Editor's TypeScript LS is not involved here,

JS/TS server is involved: it chooses the tsserver that match the project best to adapt its functionality to the project. If we force 1 tsserver version, there is a risk of misalignment between the features provided by JS/TS-LS and what the project can actually support. AFAIK, tsserver only target 1 specific version of TS/JS, it's not particularly meant to be backward compatible.

ZeeD:
@mickaelistria while I can understand your fears, my opinion is that you are conflating 2 problems and 2 responsibilities:

1)  eclipse+wwd+lsp4e...+typescript-language-server+tsserver.js , as *tools* should be internally compatible
2) the same stack should also be "compatible" with the source code edited

as I understand your fear is that this change could make 2) more complex, but I think that at the moment I am facing the problem 1: the current version of the wwd plugin, for his internal features, is not "self-consistent", but relies on tools from the user.

now, for the fact that tsserver.js compatibility is "poor" and that typescript-language-server may need to select "an appropriate version" of the tsserver.js based on the version of the source code to parse, I honestly think is a "bug" (or maybe a design choice I don't agree with) of tsserver itself, not wwd.

vrubezhny:
> I disagree. Using a newer tsserver can provide a huge feature gain to developers without having them forced to update Wild Web Developer.

Then the only option is to add a preference which allow users to select between the bundled `tsserver.js` and a custom one specified by a user.

> JS/TS server is involved: it chooses the tsserver that match the project best to adapt its functionality to the project.
In this certain case we talk about a `tsserver.js` from user project's `node_modules` folder used by JS TS Editor - this is totally wrong. What if user project isn't successfully built and looses that file at some point? What if `tsserver.js` from user project's `node_modules` is not a typescript server at all? Imho, we should never use scripts like `tsserver.js` from a user project generated output .  It's not secure and may brake the WWD functionality.

If we want to allow users to use a custom `tsserver.js` - then we must provide a configuration option that allows to do this explicitly.

ZeeD:
FWIW, even in vscode they use internal version of the tsserver:
```
[Info  - 14:12:03.559] Starting TS Server
[Info  - 14:12:03.559] Using tsserver from: /Applications/Visual Studio Code.app/Contents/Resources/app/extensions/node_modules/typescript/lib/tsserver.js
[Info  - 14:12:03.560] <syntax> Forking...
[Info  - 14:12:03.561] <syntax> Starting...
[Info  - 14:12:03.561] <semantic> Forking...
[Info  - 14:12:03.562] <semantic> Starting...
```

mickaelistria:
> the current version of the wwd plugin, for his internal features, is not "self-consistent", but relies on tools from the user.

Indeed, and it's the regular flow with TypeScript development.

> I honestly think is a "bug" (or maybe a design choice I don't agree with) of tsserver itself, not wwd.

Excatly, so it shouldn't be the responsibility of Wild Web Developer to deal with such bugs or design limitations. It's the Typescript way of things. If you don't like it, either contribute to change tsserver or don't use TypeScript, Wild Web Developer cannot mitigate such a design choice from the core language tool.

Now, back to issue #1070, we see that the generic issue is "Wild Web Developer doesn't support perfectly older TypeScript versions". Well, it's true; but it's not in your best interest nor in the best interest of the project to make backward-compatibility higher priority; so I really believe that Wild Web Developer must favor forward-compatibility and toolchain consistency (tools that match target typescript version) over backward-compatibility.
The best path of action for you is to move to newer TypeScript and everything will work better all of a sudden.

mickaelistria:
> FWIW, even in vscode they use internal version of the tsserver:

OK, that's one solid argument here. I was expecting it was doing differently.
So if you can verify it with multiple example using different TypeScript version; we can make Wild Web Developer just mimic VSCode to better align with the de-facto standard TypeScript IDE.

ZeeD:
I can confirm to you this behavior both with the "original" typescript 3.9 and with a different project that uses npm namespaces (with ts 4.7)

mickaelistria:
OK, thanks a lot @vrubezhny for the patch and @ZeeD for the comparison with VSCode to guide towards a more standard behavior.

PR #1112, opened 2023-02-28
README.md: Update installation/download links

howlger:
The [Marketplace entry](https://marketplace.eclipse.org/content/wild-web-developer-html-css-javascript-typescript-nodejs-angular-json-yaml-kubernetes-xml) seems to have changed, maybe due to the 1.0.0 release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,14 +49,14 @@ Get an Eclipse IDE for JavaScript and Web developers: download from [https://dow
 
 ### As extension into an existing Eclipse IDE installation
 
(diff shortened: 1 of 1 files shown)


mickaelistria:
Thanks. FYI, the Marketplace entry has changed because it got removed (it was using only http:// and I failed at realizing some warning I received from the Foundation were about Wild Web Developer and ignored them, until EF did what they said they would do: remove the entry) and then recreated.

PR #1123, opened 2023-03-13
Issue #1078 XML Catalog contributions from plugin are only extracted if necessary to enable includes with relative references within the plugin

### LINKED ISSUE #1078: XML catalog contribution from plugin
We develop a Eclipse plugin that contributes XML schemas to the XML catalog using the extension point `org.eclipse.wst.xml.core.catalogContributions`.

The XML schema documents are packaged with the plugin and thus use a relative URI to reference the files. What is special about the schema documents is that they use `include`-elements [https://www.w3.org/TR/xmlschema11-1/#compound-schema] to assemble multiple schema documents. The additional schema documents are as well packaged with the plugin and are referenced relative.

In contrast to the XML editor of WTP this doesn't work with the XML editor of wild web developer. The editor complains about errors in the schema document, where the number of errors correlates with the number of `include`-elements.

I could already track this down to the fact that the system catalog (`<workspace>\.metadata\.plugins\org.eclipse.wildwebdeveloper.xml\system-catalog.xml`) contains a reference to the schema document that is extracted from the plugin (somewhere like `<eclipse-installation>/configuration/org.eclipse.osgi/14/0/.cp/de/foo/schema.xsd`). Unfortunately the referenced schema documents are not extracted at all and therefore the relative references can't be resolved.

With further investigation it became evident that this is related to the use of `FileLocator.toFileURL(URL)` to resolve URIs that are not of `file`-scheme. The method is used in `resolveURI(IConfigurationElement()` of `org.eclipse.wildwebdeveloper.xml.internal.ui.preferences.XMLCatalogs`, which is called by `getWTPExtensionCatalog()`. As denoted in [JavaDoc](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/runtime/FileLocator.html#toFileURL(java.net.URL)) the contents might be extracted to a cache in the file system. This seems to be the case for URIs to plugins without further resolving.

From a users perspective the use of `include`-elements in a schema document is perfectly fine and therefore this should be supported by the XML editor.

A possible fix I can think of might be resolving the URI to a native java protocol ([FileLocator.resolve(URL)](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/runtime/FileLocator.html#resolve(java.net.URL))) and passing the schema document in its original location. The resolving should yield a jar-URI which should be perfectly fine for any java process.

Another more extensive option would be to scan the schema documents and include the referenced files in the cache as well.

blasss:
Pull Request to fix #1078

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.wildwebdeveloper.tests/META-INF/MANIFEST.MF
@@ -24,5 +24,6 @@ Require-Bundle: org.eclipse.wildwebdeveloper;bundle-version="1.0.0",
  org.eclipse.core.filesystem,
  org.eclipse.ui.editors,
  org.eclipse.ui.console;bundle-version="3.11.0",
- org.eclipse.lsp4e.debug;bundle-version="0.13.0"
+ org.eclipse.lsp4e.debug;bundle-version="0.13.0",
+ org.eclipse.wst.xml.core
 Import-Package: org.eclipse.ui.editors.text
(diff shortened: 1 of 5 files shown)


blasss:
Sorry, wrong commit user. This user will not be able to sign off ECA

PR #1131, opened 2023-03-16
[nobug] Use maven provided dependencies instead of the Orbit provided…

vrubezhny:
… ones

### DIFF
--- a/target-platform/target-platform.target
+++ b/target-platform/target-platform.target
@@ -49,7 +49,7 @@
 	<repository location="http://download.eclipse.org/cbi/updates/license/"/>
 	<unit id="org.eclipse.license.feature.group" version="0.0.0"/>
 </location>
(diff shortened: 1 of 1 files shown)


mickaelistria:
`[TestJSON.testFormatEnabled](https://ci.eclipse.org/wildwebdeveloper/job/Wildwebdeveloper/job/PR-1131/5/testReport/junit/org.eclipse.wildwebdeveloper.tests/TestJSON/testFormatEnabled/)` seems to be persistently failing on CI with this change. Is it a regression that can be reproduced locally?

vrubezhny:
@mickaelistria I believe it's a failure similar to what we had in https://github.com/eclipse/wildwebdeveloper/pull/1134 - commands are not fully  initialized at the moment they statuses, like enablement, are acquired. 

When I run the test locally from under Eclipse it succeeds, when I run it from the command line together with all other WWD tests it succeeds as well, but when I run this and only this test from the command line it fails - probably some initialization job isn't finished at the time of testing in this case.

PR #1146, opened 2023-03-21
Try LSP4E snapshots

mickaelistria:
Hoping it fixes TestESLint failure

### DIFF
--- a/target-platform/target-platform.target
+++ b/target-platform/target-platform.target
@@ -37,7 +37,7 @@
 	<unit id="org.eclipse.mylyn.wikitext_feature.feature.group" version="0.0.0"/>
 </location>
 <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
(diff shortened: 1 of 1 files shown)


target-platform/target-platform.target
```diff
@@ -37,7 +37,7 @@
 	<unit id="org.eclipse.mylyn.wikitext_feature.feature.group" version="0.0.0"/>
 </location>
 <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
-	<repository location="https://download.eclipse.org/lsp4e/releases/latest/"/>
+	<repository location="https://download.eclipse.org/lsp4e/snapshots/"/>
```
wesleybl, line 40: Wouldn't it be good to put a FIXME, to change to latest, after a new release of lsp4e?

mickaelistria: I don't think it changes anything; I'm usually the one doing releases and I don't look for "FIXME" anyway.

mickaelistria: If you're willing to contribute that, it would be highly welcome.

PR #1180, opened 2023-04-24
 Use LemMinX 0.25

vrubezhny:
LemMinX 0.25 is not API-compatible with 0.24 so extensions may not work properly.


vrubezhny:
I' m working on PRs for LemMinX-Maven (v.0.8.0 is already released), WWD and m2e-core that finally use/will use 
 and be compatible with Lemminx v.0.25.0

angelozerr:
Nice! I suggest that you consume cancel checker for hover to stop as son as possible the maven hover compute

PR #1226, opened 2023-05-26
Allow configuring maxTsServerMemory for JS/TS language server

PyvesB:
Very big projects can cause the JavaScript/Typescript language server to run out of memory and crash. There is an option available in the language server's initialisation options to override the default memory used by Node.js (see [here](https://github.com/typescript-language-server/typescript-language-server#initializationoptions)). This pull request allows configuring that option via a system property.

I've tested this change on a private repository that was previously crashing the language server. Setting `-Dorg.eclipse.wildwebdeveloper.maxTsServerMemory=16384` fixes the issue.


vrubezhny:
@PyvesB Thanks for the contribution!

PR #1233, opened 2023-06-02
Limit LSP4E dependency to version <= 0.23 due to the API--breaking ch…

vrubezhny:
…anges

### DIFF
--- a/org.eclipse.wildwebdeveloper.xml.feature/feature.xml
+++ b/org.eclipse.wildwebdeveloper.xml.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.xml.feature"
       label="%name"
-      version="1.1.0.qualifier"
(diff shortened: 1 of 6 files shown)


akurtakov:
That is not fine as that would make wwd unusable with other LSP based editors that develop against master

mickaelistria:
@vrubezhny if you can directly include a newer version of the code that works against latest release and current master of LSP4E, and restrict the ranges minimal version instead, that would be best.

vrubezhny:
> @vrubezhny if you can directly include a newer version of the code that works against latest release and current master of LSP4E, and restrict the ranges minimal version instead, that would be best.

That's the goal, but at the moment if we don't restrict LSP4E versions and users are updating to the latest LSP4E snapshot - they fail to run WWD/M2E-Core. And nothing prevents the users from such an update, plus it's 2023-06 RC1 milestone today.

vrubezhny:
Closing as not accepted

PR #1235, opened 2023-06-02
Fix API incompatobilities in latest o.e.lsp4e (v. >= 0.17)

vrubezhny:


mickaelistria (approved):
This change looks very good. Anything particular you'd like me to give more attention? If not, the sooner we merge, the better.

vrubezhny:
> This change looks very good. Anything particular you'd like me to give more attention? If not, the sooner we merge, the better.

@mickaelistria  JUnit test  [TestTypeScript.testHTMLinTSXFile() ](https://github.com/eclipse/wildwebdeveloper/blob/eaae1da1cff22d077c456ed352ad2e00d6e6ffd0/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestTypeScript.java#L76) is quite unstable failing by time out expired while trying to collect language servers for document. From one side - it's only a test that just started an editor, so the LS needs some time to initialize, so I've used 500ms timeout here, but we have a similar code collecting the LSs in [AssociateGrammarHandler.executeServerCommand](https://github.com/eclipse/wildwebdeveloper/blob/eaae1da1cff22d077c456ed352ad2e00d6e6ffd0/org.eclipse.wildwebdeveloper.xml/src/org/eclipse/wildwebdeveloper/xml/internal/commands/AssociateGrammarHandler.java#L97) - I've put 100ms timeout here - in theory at least one LS is to be initialized and active at this point, but I still in doubt on what should be the correct timeout value here?

mickaelistria:
This test is working on the Jenkins build for this PR, so I woudn't worry about it. If it is the only concern, then I believe this PR is good to merge now.

PR #1308, opened 2023-08-14
Allow using TypeScript version specified by project

PyvesB:
Until now, the TypeScript version was systematically being overridden to point to the one packaged with Eclipse. This meant that if your project relied on a different version of TypeScript (as specified in `package.json` for example), there was a risk of incompatibilities. Even worse, custom type definitions brought in third party dependencies were not resolved properly, for example in the case of React.js:
Screenshot 2023-08-14 at 12 17 04

VSCode allows switching between either its own version of TypeScript, or the one specified by the project, as described here:
https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript

This PR adds similar support to Eclipse, via a preference page option. The default is the same as before, i.e. TypeScript packaged with Eclipse, so the behaviour for current Eclipse users will not change.

Tested on a project locally, all type definition resolution errors go away, and you can see the version that is being picked up by referring to the log messages produced by the language server at startup (as documented [here](https://github.com/typescript-language-server/typescript-language-server#typescript-version-notification)).

With the Eclipse version:
```
Typescript Server version info: {version=5.0.4, source=user-setting}
```

With the Project version:
```
Typescript Server version info: {version=5.0.2-sdk, source=workspace}
```

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/ui/preferences/JSTSPreferenceServerConstants.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/ui/preferences/JSTSPreferenceServerConstants.java
@@ -1,0 +1,38 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Pierre-Yves Bigourdan - initial implementation
+ *******************************************************************************/
+package org.eclipse.wildwebdeveloper.jsts.ui.preferences;
+
+import org.eclipse.jface.preference.IPreferenceStore;
+import org.eclipse.wildwebdeveloper.Activator;
+
+/**
+ * JS/TS preference server constants.
+ *
+ */
+public class JSTSPreferenceServerConstants {
+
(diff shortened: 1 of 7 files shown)


org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/ui/preferences/JSTSPreferenceServerConstants.java
```diff
@@ -0,0 +1,38 @@
+/*******************************************************************************
```
vrubezhny, line 1: Why not to merge this class into https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/blob/master/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/ui/preferences/typescript/TypeScriptPreferenceServerConstants.java ?

vrubezhny (commented):
Please see the inline comment.

org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/jsts/ui/preferences/JSTSPreferenceServerConstants.java
PyvesB: Are you also suggesting to move the preference to the TypeScript subsection in the UI? My rationale was that this preference effectively impacted the language server as a whole, so I put it somewhere top level.

vrubezhny: No. I think you're right. Let's live it as it is - to be a common setting for the language server.

vrubezhny (approved):
Looks good to me. Thanks!

PR #1312, opened 2023-08-19
Add snakeyaml dependency to wildwebdeveloper repository

HannesWell:
This avoids the need to add to sankeyaml in downstream consumer (e.g. m2e) via Orbit, if they don't use the SimRel repo (which I was told is not recommended).


mickaelistria:
Thanks.

Issue #1316, opened 2023-08-23
Vue.JS tests failures


vrubezhny:
Vue.JS JUnit Tests have started to fail permanently:

```
Running org.eclipse.wildwebdeveloper.tests.TestVue
...
[/home/jeremy/projects/wildwebdeveloper/org.eclipse.wildwebdeveloper.tests/target/work/.node/node-v18.17.1-linux-x64/bin/npm, install, --no-bin-links, --ignore-scripts]
Error Stream: >>>
npm WARN deprecated stable@0.1.8: Modern JS already guarantees Array#sort() is a stable sort, so this library is deprecated. See the compatibility table on MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#browser_compatibility
npm WARN deprecated consolidate@0.15.1: Please upgrade to consolidate v1.0.0+ as it has been modernized with several long-awaited fixes implemented. Maintenance is supported by Forward Email at https://forwardemail.net ; follow/watch https://github.com/ladjs/consolidate for updates and release changelog
<<<
Output Stream: >>>

added 952 packages, and audited 953 packages in 31s

110 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
<<<
Aug 23, 2023 2:02:49 PM org.eclipse.tm4e.core.registry.Registry _doLoadSingleGrammar
WARNING: No grammar source for scope [text.html.markdown]
Aug 23, 2023 2:02:49 PM org.eclipse.tm4e.core.registry.Registry _doLoadSingleGrammar
WARNING: No grammar source for scope [text.html.derivative]
Aug 23, 2023 2:02:49 PM org.eclipse.tm4e.core.registry.Registry _doLoadSingleGrammar
WARNING: No grammar source for scope [text.pug]
Aug 23, 2023 2:02:49 PM org.eclipse.tm4e.core.registry.Registry _doLoadSingleGrammar
WARNING: No grammar source for scope [source.stylus]
Aug 23, 2023 2:02:49 PM org.eclipse.tm4e.core.registry.Registry _doLoadSingleGrammar
WARNING: No grammar source for scope [source.postcss]
Aug 23, 2023 2:02:49 PM org.eclipse.tm4e.core.registry.Registry _doLoadSingleGrammar
WARNING: No grammar source for scope [source.sass]
[...]

PR #1317, opened 2023-08-23
Use latest TM4E release from TM4E own repo instead of the Eclipse rel…

vrubezhny:
…eases


vrubezhny:
The Vue.js test failures (reported in #1316) are not related to this PR

PR #1322, opened 2023-08-26
[tests] Ignore 'DirectoryNotEmptyException' exception...

vrubezhny:
... as it creates too much noice it the log while running JUnit tests


vrubezhny:
The Vue.js test failures (reported in #1316) are not related to this PR

PR #1328, opened 2023-09-08
#1327 A "Chrome Debug" session can now be started even when a chrome window is already open

### LINKED ISSUE #1327: Starting a "Chrome Debug" session does not work when chrome is already open
# Steps to reproduce # 
Exact same steps as described in #1294 except before starting the "Chrome Debug" session open a chrome window. 

Instead of opening a new chrome instance, a new empty tab is opened in the running chrome window. wwd is not apple to connect to this chrome instance.

# Environment # 
Eclipse 2023-06
wwd 1.3.0
I tested this both on Windows 11 and Ubuntu 20

# Analysis # 
Since the chrome instance was not started with remote debugging enabled, wwd is not able to connect. 

# Workaround # 
As already mentioned in #1294: Close all chrome windows before starting a "Chrom Debug" session. 
This workaround is problematic, because it might interrupt a developer's normal workflow. 

# Related issues # 
#927 is probably related. The difference in #927 is that the error occurs when starting a "Running Chrome Debug Instance" debug session and not a "Chrome Debug" session.

Si-So:
# Rationale # 
The "vscode-chrome-debug" debugger can either be instructed to start chrome from a specific path or it can try to locate chrome itself. See here: https://github.com/microsoft/vscode-chrome-debug/blob/master/src/chromeDebugAdapter.ts
If going for the first option the `runtimeExecutable` parameter must be specified. This is what wwd currently does and what causes #1327

In contrast, here is a typical `launch.json` from VSCode: 
```
{
    // Use IntelliSense to learn about possible attributes.
    // Hover to view descriptions of existing attributes.
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        {
            "type": "chrome",
            "request": "launch",
            "name": "Launch Chrome against localhost",
            "url": "http://localhost:4200/",
            "webRoot": "${workspaceFolder}/src",
            "sourceMaps":true
        }
    ]
}
```
Instead of using the `runtimeExecutable` parameter, VSCode uses the `type` and `launch` parameters. When the `type` and `launch` parameters are used "vscode-chrome-debug" locates the chrome executable itself. This fixes #1327 

I am not sure why it fixes #1327, but it does. 

# Appendix # 
This only fixes "Chrome Debug" and _not_ "Running Chrome Debug Instance". I suspect that "Running Chrome Debug Instance" can be fixed in the same way. After also fixing "Running Chrome Debug Instance" the class `ChromeExecutableTab` probably becomes obsolete. The method `findChromeLocation` in `ChromeRunDAPDebugDelegate` can probably also be deleted.

### DIFF
--- a/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
+++ b/org.eclipse.wildwebdeveloper/src/org/eclipse/wildwebdeveloper/debug/chrome/ChromeRunDAPDebugDelegate.java
@@ -40,8 +40,6 @@ import org.eclipse.ui.internal.browser.IBrowserDescriptor;
 import org.eclipse.wildwebdeveloper.Activator;
 import org.eclipse.wildwebdeveloper.debug.AbstractHTMLDebugDelegate;
 import org.eclipse.wildwebdeveloper.debug.LaunchConstants;
-import org.eclipse.wildwebdeveloper.debug.MessageUtils;
-import org.eclipse.wildwebdeveloper.debug.Messages;
 
 import com.google.gson.JsonObject;
 
@@ -54,6 +52,10 @@ public class ChromeRunDAPDebugDelegate extends AbstractHTMLDebugDelegate {
 	public static final String RUNTIME_EXECUTABLE = "runtimeExecutable";
 	public static final String URL = "url";
 	private static final String SOURCE_MAPS = "sourceMaps";
+	public static final String TYPE = "type";
+	public static final String CHROME = "chrome";
+	public static final String REQUEST = "request";
+	public static final String LAUNCH = "launch";
 
 	@Override
 	public void launch(ILaunchConfiguration configuration, String mode, ILaunch launch, IProgressMonitor monitor)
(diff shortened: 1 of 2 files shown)


mickaelistria:
Indeed, it looks like `vscode-chrome-debug` plugin does a better job at finding chrome that Wild Web Developer does, so it's a good thing it we can rely on it more.

mickaelistria:
Thank you!

Si-So:
Thank you for merging :)

Before merging my PR: I also found it quite strange that you even _could_ select a browser in "Chrome Debug".

mickaelistria:
> Before merging my PR: I also found it quite strange that you even could select a browser in "Chrome Debug".

At this stage, you currently seem like the most expert contributor regarding Chrome debug integration. So at least I, and probably others, trust your judgement here and would welcome any PR that in your opinion removes some "strangeness".

mickaelistria:
We don't plan much, this is a bit "on-demand" and according to availability of committers.
However, let me turn the question upside-down: why would you need a release?

Si-So:
Yeah I noticed that your release are "on-demand". 

I would like to have a release because this makes it easier for other employees from my company to get my changes. 

Manually deploying locally built plugin-jars is harder than saying: "Please update this plugin from the official repo".

mickaelistria:
There is an official repo with snapshots which you can share with your colleagues to install a build including your latest contributions: https://download.eclipse.org/wildwebdeveloper/snapshots/

Si-So:
Oh, I didn't know that. Thank you :)

PR #1341, opened 2023-09-24
Compatibility with WTP SSE

zulus:
Low priority for all WWD content-types


mickaelistria:
Why a lower priority?

zulus:
WTP SSE based editors (CSS, XML, HTML and PHP), require special IDocument implementation (IStructuredDocument),  build via deprecated "org.eclipse.core.filebuffers.documentCreation" which is based on contentType.

Without this all SSE editors stops working after WWD installation. As an alternative WTP can set his content type priority as high, or finally replace IDocumentFactory by for example ForwardingDocumentProvider ;)

mickaelistria:
Would it be possible to fix the `org.eclipse.core.filebuffers.documentCreation` so it takes into account all content-types and matches the first one that has an association instead of just looking at the first content-type?

zulus:
This ex point is deprecated, I'm afraid none care ;)

Maybe on WTP side, someone will be able to just write IDocumentProvider designed for StructuredSourceEditor and stop using this ex point

mickaelistria (approved):
I don't think this is a harmful change anyway, so feel free to merge as soon as CI is happy with it.

StefanLindner:
Now everything works fine. Thank you so much!

PR #1342, opened 2023-09-24
Fix #1247 - Maven build use node from PATH on mac os

### LINKED ISSUE #1247: Maven build use node from PATH on mac os
For now this code always return true on MAC: https://github.com/eclipse/wildwebdeveloper/blob/e52e0990f609f3dcbac5648a4bd012ecf9d753a5

Due this process builder always receiving NPM absolute path only, so use node from PATH (if exists) or fail (`/bin/env node`)

I think line from above should be just dropped, npm should always be prefixed by built-in node (if exists)

zulus:
Introduce NodeJSManager helpers to always produce ProcessBuilder with commands:

node_executable npm_location rest commands

these command is also used in NpmLaunchDelegate!

### DIFF
--- a/org.eclipse.wildwebdeveloper.embedder.node.feature/feature.xml
+++ b/org.eclipse.wildwebdeveloper.embedder.node.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.wildwebdeveloper.embedder.node.feature"
       label="%name"
-      version="1.0.6.qualifier"
+      version="1.0.7.qualifier"
       provider-name="Eclipse Wild Web Developer project"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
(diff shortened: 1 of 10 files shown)


vrubezhny:
@zulus I don't get how does it help to use embedded Node.js instead of a version that is found and used when `npm` (or `NPM.CMD`) is executed with 
```
#!/usr/bin/env node
````

zulus:
> @zulus I don't get how does it help to use embedded Node.js instead of a version that is found and used when `npm` (or `NPM.CMD`) is executed with
> 
> ```
> #!/usr/bin/env node
> ```

Because rather than just run:
`$: npm install something`

Based on executable location search for  `cli.js` (or `npm-cli.js`) and execute:

`$ path_to_npm_builtin_or_from_property path_to_cli.js install something`

zulus:
Just check console log for tests (MacOS, Windows, Linux)

vrubezhny:
Won't this `path_to_cli.js`  still be executed with `#!/usr/bin/env node`?

> Just check console log for tests (MacOS, Windows, Linux)

I had Node.js 20 configured for this purpose on MacOS GitHub build action :)

zulus:
After this path no both will be directly from embedder

vrubezhny (approved):
Looks good to me

PR #1344, opened 2023-09-24
Fix github workflows

zulus:
Test PR, I'm trying to stabilize tests

### DIFF
--- a/.github/workflows/windowsPR.yml
+++ b/.github/workflows/windowsPR.yml
@@ -24,6 +24,6 @@ jobs:
         maven-version: 3.9.2
     - uses: actions/setup-node@v3
       with:
-        node-version: '16'
+        node-version: '18'
     - name: Build with Maven
       run: mvn -B clean verify
--- a/org.eclipse.wildwebdeveloper/package.json
+++ b/org.eclipse.wildwebdeveloper/package.json
@@ -9,6 +9,7 @@
     "yaml-language-server": "1.14.0",
     "vscode-languageserver-protocol": "3.17.4-next.3",
     "vscode-languageserver-types": "3.17.4-next.2",
+    "vscode-languageserver": "8.1.0",
     "vscode-css-languageserver": "file:target/vscode-css-languageserver-1.0.0.tgz",
     "vscode-html-languageserver": "file:target/vscode-html-languageserver-1.0.0.tgz",
     "vscode-json-languageserver": "file:target/vscode-json-languageserver-1.3.4.tgz",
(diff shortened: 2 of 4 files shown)


.github/workflows/windowsPR.yml
```diff
@@ -24,6 +24,6 @@ jobs:
         maven-version: 3.9.2
     - uses: actions/setup-node@v3
       with:
-        node-version: '16'
+        node-version: '18'
```
zulus, line 27: Make sure we use same version as in embedder

org.eclipse.wildwebdeveloper/package.json
```diff
@@ -9,6 +9,7 @@
     "yaml-language-server": "1.14.0",
     "vscode-languageserver-protocol": "3.17.4-next.3",
     "vscode-languageserver-types": "3.17.4-next.2",
+    "vscode-languageserver": "9.0.0",
     "vscode-css-languageserver": "file:target/vscode-css-languageserver-1.0.0.tgz",
```
zulus, line 13: After this vscode-languageserver is in root so nested node_modules is not required

org.eclipse.wildwebdeveloper/pom.xml
```diff
@@ -204,6 +204,8 @@
 								<arg>--only=prod</arg>
 								<arg>--no-bin-links</arg>
 								<arg>--ignore-scripts</arg>
+								<arg>--install-strategy=hoisted</arg>
+								<arg>--force</arg>
```
zulus, line 208: Make sure dependencies are always fresh, without this tests fail on github but pass jenkins

zulus: Force also will ignore platform requirements, so will be possible to add fsevents for mac users ;)

org.eclipse.wildwebdeveloper/package.json
zulus: I downgraded to 8.2 (9.0 have just 5 days)

zulus:
For now all tests pass, but I'm afraid sooner or later multiple "package.json" will be required to avoid too long resource paths

mickaelistria:
Thank you!

vrubezhny:
Long resource path will highly likely appear until vscode updates its language servers support, see:
- [Iteration Plan for September 2023 #192822](https://github.com/microsoft/vscode/issues/192822)
- [Prepare for 3.17.4 release #1314](https://github.com/microsoft/vscode-languageserver-node/issues/1314)

zulus:
I hope. For now yaml and angular embed and use vscode-languageserver 7.0.0 (protocol 3.16.0), vue rely on vscode-languageserver 8.1.0

PR #1372, opened 2023-10-19
Use "outputCapture": "std" for Node launches

### LINKED ISSUE #1365: Node program output no longer shown in Console view
With WWD 1.3.1.202310101531, running or debugging a JS/TS file no longer has its standard output/error shown in the Console view.

All that's shown in the Console view is this:
```
/<path-to-eclipse>/.node/node-v18.18.0-linux-x64/bin/node ./<path-to-launched-file>
```
followed by
```
Process exited with code 1
```
once the program exits.

There is also another Console view created for the debug adapter, but that isn't for program output.

Notably, running the exact same command (and in the same working directory as what the Eclipse launch uses) from a separate terminal does show program output, so it's not as though there's an issue with the plugin's embedded Node.

AndrewFerr:
Fixes #1365


mickaelistria:
thank you!

PR #1402, opened 2023-11-16
[nobug] Bump version towards 1.3.2

vrubezhny:


mickaelistria:
What is the change that requires bumping the version? We should only bump versions when it's needed.

vrubezhny:
> What is the change that requires bumping the version? We should only bump versions when it's needed.

The following merge of https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/1382 which is approved by the IP Team

mickaelistria:
OK, cool!

Issue #1403, opened 2023-11-16
fsevents.node is not signed


jonahgraham:
Wild Web Developer contains some unsigned code which prevents products from being notarized. This file is not signed: `org.eclipse.wildwebdeveloper_1.1.2.202311151451/node_modules/fsevents/fsevents.node`

The full error I see is as follows, with similar errors for other EPPs.

```json
{
  "logFormatVersion": 1,
  "jobId": "8d36c15c-b650-4242-b817-3bb211054150",
  "status": "Invalid",
  "statusSummary": "Archive contains critical validation errors",
  "statusCode": 4000,
  "archiveFilename": "eclipse-php-2023-12-M3-macosx-cocoa-x86_64-11621221555228008492.dmg",
  "uploadDate": "2023-11-16T18:28:47.503Z",
  "sha256": "13d8a69e550050f48a92a8537e986c996218c5066519d6f7485cfd77bf29c34e",
  "ticketContents": null,
  "issues": [
    {
      "severity": "error",
      "code": null,
      "path": "eclipse-php-2023-12-M3-macosx-cocoa-x86_64-11621221555228008492.dmg/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_1.1.2.202311151451/node_modules/fsevents/fsevents.node",
      "message": "The binary is not signed.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087721",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "eclipse-php-2023-12-M3-macosx-cocoa-x86_64-11621221555228008492.dmg/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_1.1.2.202311151451/node_modules/fsevents/fsevents.node",
      "message": "The signature does not include a secure timestamp.",
      "docUrl": "https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/resolving_common_notarization_issues#3087733",
      "architecture": "x86_64"
    },
    {
      "severity": "error",
      "code": null,
      "path": "eclipse-php-2023-12-M3-macosx-cocoa-x86_64-11621221555228008492.dmg/Eclipse.app/Contents/Eclipse/plugins/org.eclipse.wildwebdeveloper_1.1.2.202311151451/node_modules/fsevents/fsevents.node",
[...]

PR #1404, opened 2023-11-17
[build] Sign '*.node' files

vrubezhny:
This PR has the goal to resolve issue #1403 by implementing the steps that are performed in Orbit to sign Mac '*.node' NPM modules for the WWD build.


vrubezhny:
Mac signing breaks the WWD Jarsigner's signature as it changes and re-packs the already signed jar

vrubezhny:
This time o.e.wwd plugin's jar appears to be correctly signed, however there is still some unsigned contents:
image

However I'm not sure if this is to be solved in WWD

vrubezhny:
@jonahgraham Can you try building an EPP  using the following  PR's updated site: https://ci.eclipse.org/wildwebdeveloper/job/Wildwebdeveloper/job/PR-1404/6/artifact/repository/target/repository/ (or any later one if exists by the time)?

mickaelistria:
> This time o.e.wwd plugin's jar appears to be correctly signed, however there is still some unsigned contents:

Those are supposed to be signed by PGP, after the "assemble-repository" step.

vrubezhny:
> > This time o.e.wwd plugin's jar appears to be correctly signed, however there is still some unsigned contents:
> 
> Those are supposed to be signed by PGP, after the "assemble-repository" step.

Then I expects we're OK with this (Just need to polish pom and remove not needed profile call added to Jenkinsfile)

PR #1246, opened 2023-06-09
Update CSS language configuration and grammars to the latest ones fro…

vrubezhny:
…m VSCode 1.79.0

PR #1429, opened 2023-12-12
fix: misspelled "#property-values" include

sebthom:
The file does not define a `property-value` but not a `property-values` rule - which is correctly referenced in all other include directives.

### DIFF
--- a/org.eclipse.wildwebdeveloper/grammars/css/css.tmLanguage.json
+++ b/org.eclipse.wildwebdeveloper/grammars/css/css.tmLanguage.json
@@ -1057,7 +1057,7 @@
 							"name": "support.constant.text-direction.css"
 						},
 						{
-							"include": "#property-value"
+							"include": "#property-values"
 						}
 					]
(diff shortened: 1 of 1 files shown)


mickaelistria:
If I'm not mistaken, this file is directly copied from VSCode. Can you please check whether the issue is present in VSCode file too?

sebthom:
The issue is not present in upstream anymore. The CSS file is actually outdated and there is another draft PR https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/1246 to update all grammar files but it is stalled for already six months. The change of my PR however is trivial and safe/quick to merge whereas #1246 probably needs more work/time (IP review etc).

akurtakov:
There is no longer css grammar anymore with https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/commit/796ae273669719c5deb208dcf045ab5d23cc8cbf . The one from tm4e is used.

PR #1432, opened 2023-12-21
Fix preferences layout

Bananeweizen:
before
grafik

and after
grafik

Checkboxes (on Windows) don't have multiline text, at least not without additional magic, therefore the newline is invalid there.


mickaelistria:
It looks like version needs to be bumped.

Bananeweizen:
> It looks like version needs to be bumped.

I have trouble building locally, so I might need some more attempts to fix it. I'll repeat adding feature changes, until it's green. :)

mickaelistria:
Thanks!

PR #1433, opened 2023-12-21
Use target as file instead of as module

Bananeweizen:
This is possible since quite a while in Tycho. Main benefit is that you can build single maven modules with this approach, while the old artifact style allows only building the complete reactor (or requires installing the target artifact).

The maven nature and POM can be removed from the target directory.


mickaelistria:
thanks!

PR #1434, opened 2023-12-21
Upgrade Tycho

Bananeweizen:
* upgrade minimum Maven version, Tycho 4 requires 3.9
* upgrade Tycho itself. why didn't dependabot do this?
* remove invalid resolver configuration


mickaelistria:
Thanks!

PR #1442, opened 2024-01-15
Include newer snakeyaml (from TM4E repo)

mickaelistria:


mickaelistria:
Not worse than master, and unblocks PR builds. Let's merge.

PR #1443, opened 2024-01-15
Revert "Bump typescript-language-server from 4.0.0 to 4.2.0 in /org.eclipse.wildwebdeveloper"

mickaelistria:
Reverts eclipse-wildwebdeveloper/wildwebdeveloper#1428
As this cause some new test failures.


mickaelistria:
@vrubezhny This seems to fix the failing tests. Should we apply it? I don't think I'll personally have the opportunity to debug it further, so unless you can dedicate time to make typescript-ls 4.2 or 4.3 work, it's safer to revert to a state that it known to work well.

PR #1484, opened 2024-03-05
GitHub Actions, build on push and pr on all OS

jonahgraham:
I'm not sure if there was a reason that PRs weren't being built on GHA with all platforms, so I provide this updated builder.

There is a warning (presumably preexisting and not affecting JIPP):

`
Warning:  Using JavaSE-21 to fulfill requested profile of JavaSE-17. This might lead to faulty dependency resolution, consider defining a suitable JDK in the toolchains.xml.
`


jonahgraham:
My primary purpose was to make sure build still worked with the updated LSP4E/LSP4J combination as there were some breaking changes in LSP4J's latest release. Since there were no github actions for "push" I thought I would create and contribute them.

jonahgraham:
Looks like more to do here - and since the Jenkins build passed (except a test failure) I am abandoning this.

PR #1531, opened 2024-04-23
Let Tycho handle metadata with default

akurtakov:
The config for signing only makes it suspect for
https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/issues/1505 .


akurtakov:
As only master build is signed, push directly for faster test.

Issue #1505, opened 2024-03-26
Inconsistent sha512 sum in snapshot site


mickaelistria:
### Discussed in https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/discussions/1485

<div type='discussions-op-text'>

<sup>Originally posted by **jcompagner** March  6, 2024</sup>
when i want to update to:

Wild Web Developer HTML, CSS, JSON, Yaml, JavaScript, TypeScript, Node tools	1.3.4.202402272001

i get:

An error occurred while collecting items to be installed
session context was:(profile=SDKProfile, phase=org.eclipse.equinox.internal.p2.engine.phases.Collect, operand=, action=).
Problems downloading artifact: osgi.bundle,org.eclipse.wildwebdeveloper,1.1.5.202402272001.
SHA-512 hash is not as expected. Expected: 9e75fc48a51ffa9635c01e72dc02c5f41193c9c8a3efe9a273c4d069240814bf664114734de7cfa9ff514ba3841336d2c4343390340262ad6c3c01674dd0e81e and found d75df3e7cedce2d8b55b90aa3343217fac24c922041310e3f152cc0f2ad32781019b766e052a676251e25fb43847d3993183af1de4c1f0a898779cc8cd46d12e.

anybody else has that problem?

and i like to update, to see where the problem is that i currently have in "ts" editor. 
If i copy/paste then i get extra spaces before it (and the copy really doesn't have that, paste in another editor works fine)
</div>

jcompagner:
i tried the one that i guess is from today:

An error occurred while collecting items to be installed
session context was:(profile=SDKProfile, phase=org.eclipse.equinox.internal.p2.engine.phases.Collect, operand=, action=).
Problems downloading artifact: osgi.bundle,org.eclipse.wildwebdeveloper,1.1.5.202403260750.
SHA-512 hash is not as expected. Expected: b6ef06399fe41e1dcdb4a341b095df3b6bc92c0536b8414f681897893094fe9afd6c0c60e2b3c0aaba65dd67c64e88bee9d3f57c9e6a406142af932961a22caf and found 19c3c341e0e311612689d46e8c49e00c58d1ca83772338332be6bee110eb5d95654bac9c19fdabb789f0c674f29f205b8c01a2bec3caae1a1d33c8a6ea2ef046.

mickaelistria:
So it could be the signatures that mess up the sha512sum.
[...]

PR #1540, opened 2024-04-23
Do not alter already signed jar

mickaelistria:
Fixes https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/issues/1505


mickaelistria:
I verified and this fixes #1505 and so I think it should be merged ASAP. However how can we verify is doesn't reintroduce #1403 ? @vrubezhny @jonahgraham any hint on how to verify the macos signature stuff?

PR #1578, opened 2024-05-27
Fix generics warnings

akurtakov:


akurtakov:
License check failure is yet another transitive dep one from npm land that is in master too.

PR #1580, opened 2024-05-28
Update maven plugins

akurtakov:


akurtakov:
Seems to be unstable tests.

PR #1628, opened 2024-07-18
update lemminx to 0.28.0

martinlippert:
The lemminx language server in version 0.27.0 throws a nasty error in some cases (see https://github.com/eclipse/lemminx/issues/1653) that has been fixed on 0.28.0, so updating to that version would be awesome.

### DIFF
--- a/org.eclipse.wildwebdeveloper.xml/pom.xml
+++ b/org.eclipse.wildwebdeveloper.xml/pom.xml
@@ -40,7 +40,7 @@
 									<groupId>org.eclipse.lemminx</groupId>
 									<artifactId>org.eclipse.lemminx</artifactId>
 									<!-- Bumping to version with API breakage needs to bump bundle version at least by +0.1.0 -->
-									<version>0.27.0</version>
+									<version>0.28.0</version>
 									<!-- classifier:uber includes all dependencies -->
 									<classifier>uber</classifier>
 								</artifactItem>


mickaelistria:
The eslint failures are not related, let's merge.

martinlippert:
Awesome, thanks @mickaelistria. Any idea when a new release build might be around that includes this upgrade? Any plans?

mickaelistria:
I think we can release as soon as the 4 failing ESLint tests are resolved. However, I'm not aware of anyone looking at them nor planning to look at them on short term. If you can spend a bit of time investigating those, that would be helpful.

martinlippert:
> I think we can release as soon as the 4 failing ESLint tests are resolved. However, I'm not aware of anyone looking at them nor planning to look at them on short term. If you can spend a bit of time investigating those, that would be helpful.

I will most likely not be able to spend time on this myself at the moment (vacation coming up), but I will ask someone from the team to support here.

mickaelistria:
We have reverted the change that was causing the 4 failures, so no further work is needed here.
The current issue that prevents from a release is some IP approval for some NPM dependencies (I don't remember which ones and the build log isn't verbose enough to figure it out)

laeubi:
I have one question, before it seems m2e has provided the lemminx version via extension point

```
<plugin>
   <extension
         point="org.eclipse.wildwebdeveloper.xml.lemminxExtension">
      <jar
            path="...the jar...">
      </jar>
      <classpathExtensionProvider
            provider="org.eclipse.m2e.editor.lemminx.MavenRuntimeClasspathProvider">
      </classpathExtensionProvider>
      <initializationOptionsProvider
            provider="org.eclipse.m2e.editor.lemminx.InitializationOptionsProvider">
      </initializationOptionsProvider>
   </extension>
</plugin>
```

but this seem to no longer work, what ever I tried out I end up with:

```
INFORMATION: Initializing XML Language server
LemMinX Server info:
 - Version : 0.28.0
 - Java : /opt/java/jdk-21
 - VM Version : 21
 - Git 9c3a0f1 - [maven-release-plugin] prepare release 0.28.0
```

if I **delete** the jar then the server seem to fail to start, but beside from that the jar needs to be there it somehow always seem to load the "old" (current released) version.

laeubi:
I just found out that this is only for lemminx-maven (not lemminx itself) ... What seem to work is find out where Equinox extracts the bundle content, then replace the `org.eclipse.lemminx-uber.jar` there so quite inconvenient...

PR #1721, opened 2024-11-18
Use lambda methods of DisplayHelper

akurtakov:
Get rid of all the anonymous classes generated.


akurtakov:
clearlydefined seems to be down causing licensecheck tool to fail. As there is no change in deps in this PR - merging.

PR #1783, opened 2025-02-06
Update to Tycho 4.0.11

laeubi:


laeubi:
NPM should not be relevant for this

akurtakov:
> NPM should not be relevant for this

It isn't but I'm using the chance to get cqs opened if they aren't already.

PR #1787, opened 2025-02-08
Add update target platform workflow and enable tycho-build-extension

laeubi:
This adds a workflow that automatically check for updates in the target
platform. 

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/.github/workflows#updatetargetyml

Additionally it enables the tycho-build extension and removes a no longer needed `pom.xml` to ease version updates in the future.


mickaelistria:
Let's try it.

PR #1817, opened 2025-03-18
Update eslint vscode to 3.0.10

akurtakov:


akurtakov:
I have changed jobs and have zero time for that for this now. It would have to be someone else that works on it.

vrubezhny:
> I checked this once and I remember that the version 3 server needs to receive an extra parameter in the calls. I don't remember exactly which one.

@wesleybl Please feel free to come back here commenting if you eventually remember or find that "extra parameter" out once again. Any help is appreciated.

PR #1845, opened 2025-05-04
Specify EPL-20 Bundle-License headers

merks:


merks:
These follow this structure:

https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module-bundle-license

Technically the about.html files are missing.

merks:
FYI, it looks like lemminx was recovered from older storage and is missing newer content. I've opened this issue:

https://github.com/eclipse-lemminx/lemminx/issues/1741

merks:
Thank you for moving this along!

vrubezhny (approved):
Looks good to me.  Thanks Ed!

PR #1856, opened 2025-05-28
Allow local Maven builds with dirty Git working tree

sebthom:
We use the same config in LSP4E and TM4E


vrubezhny (approved):
Looks good to me. Thanks!

PR #1552, opened 2024-05-09
Switch css/less/scss to tm4e grammar/language config

akurtakov:
Simplifies wildwebdeveloper development as the project will get the updates from tm4e without the need for CQs and etc.

PR #1860, opened 2025-05-29
Use JSON grammar/config from TM4E

sebthom:
Reduce duplication/maintenance efforts and potential conflicts.

Similar to #1552 #1553 #1819 #1820 #1821


akurtakov:
Thanks!

PR #1968, opened 2025-11-17
ci: don't run license check if dependencies don't change

sebthom:
This avoids unnecessary calls to the ClearlyDefined API which constantly fails with rate limit errors.

### DIFF
--- a/.github/workflows/licensecheck.yml
+++ b/.github/workflows/licensecheck.yml
@@ -1,20 +1,32 @@
-# This workflow will build a Java project with Maven
-# For more information see: https://help.github.com/actions/language-and-framework-guides/building-and-testing-java-with-maven
-
+# https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax
 name: License check
 
-on:
+on:  # https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows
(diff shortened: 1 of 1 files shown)


vrubezhny:
Please be careful with it, because some transitive dependencies may change their version significantly between two WWD builds even if our own package.json/pom.xml doesn't change.

mickaelistria:
> Btw, it was an advice from Eclipse CBI to not run the license check if dependencies don't change.

Checking whether dependencies changed is actually a more complex work than calling CBI dash-analysis (which is more or less just querying a database/bugtracker). It would require, additionally to a local build, fetching the last published snapshot of wildwebdeveloper, extra the package-lock.json out of it and compare the meaningful parts of it with what was just built.
I would expect the CBI dash service to be implemented well enough to support being "free" when querying some already used dependencies. If not, then the problem is IMO more to fix on CBI end than on every consumer having to set up some dependency diff mechanism.

akurtakov:
So do we want this one or no?

sebthom:
I personally want lock files :-) So we don't have surprises like two weeks ago when I wanted to create a release and suddenly one of the not-locked transitive dependencies got updated and we had to wait for manual license approval and nearly missed the simrel deadline.

PR #2128, opened 2026-06-10
Stop GHA verification on MacOS Intel

akurtakov:


akurtakov:
Merge without waiting for jenkins as this is gha only change.

PR #2137, opened 2026-06-19
Comment on PR and fail run when NPM license review is required

akurtakov:
Previously the license check only printed a message and exited successfully when unvetted NPM dependencies were found. Now the workflow parses the dash-licenses review summary, posts the list of dependencies requiring review as a PR comment, and marks the run as failed.


akurtakov:
Merge to test it on another PR.

PR #2143, opened 2026-06-20
Releng updates

merks:
- Improve the display details in the summary.
- Remove ls -sailR from promotion.
- Update org.eclipse.lemminx to 0.31.2.


akurtakov (approved):
LGTM

PR #2144, opened 2026-06-20
Contribute initial Oomph project setup and configuration

merks:


merks:
Has no build impact.  Merging immediately.

PR #2153, opened 2026-06-24
Eliminate unchecked cast in o.e.w.ui.preferences.Settings.findSettings

merks:


merks:
The buidl passed but the deploy of master appears blocked  Merging anyway.

PR #2156, opened 2026-06-25
Use Files.isDirectory/isWritable/isExecutable to determine permissions

merks:
- Fix NodeJSManager.probeDirectoryForInstallation(File) to use java.nio.Files operations which is more reliable.
- File.canWrite is not reliable on Windows with its complex access-list-based permissions. Also File.canExecute is not meaningful for a folder and is intended to indicate if a file (not a directory) is executable.

https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/issues/2106


akurtakov (approved):
LGTM

PR #2196, opened 2026-08-13
Make the JSON server's computation limit configurable

vogella:
The VSCode JSON language server computes document symbols and folding ranges for the whole document unless the client configures a limit, and Wild Web Developer never sent any configuration at all. On a 100,000 line JSON file that means 20,001 symbols and folding ranges per request, and with the Outline view open the UI thread was busy for two to three seconds per keystroke, which makes such files impractical to edit.

This adds a JSON preference page with the maximum number of computed items, defaulting to 5000 just like VSCode's `json.maxItemsComputed` and the existing YAML preference, and 0 to switch the limit off. In the same measurement the editor now spends about 23 ms of UI thread time per keystroke with the Outline view open, which is what it already cost with the Outline closed. The limit is sent through the usual preference machinery, so changing it takes effect without restarting the language server.

Two entries that only restate defaults have to be sent along, because this server re-reads all of its settings from every notification: without `json.validate.enable` it silently stops validating, and without `http.proxyStrictSSL` it stops verifying certificates when downloading remote schemas.

Note that outlines of files with more than 5000 symbols are now truncated, exactly as they are in VSCode and as YAML files already are here.

### DIFF
--- a/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestJSONResultLimitPreference.java
+++ b/org.eclipse.wildwebdeveloper.tests/src/org/eclipse/wildwebdeveloper/tests/TestJSONResultLimitPreference.java
@@ -1,0 +1,113 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Vogella GmbH and others.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *  Lars Vogel (Vogella GmbH) - initial implementation
+ *******************************************************************************/
+package org.eclipse.wildwebdeveloper.tests;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.eclipse.core.resources.IFile;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.lsp4e.LSPEclipseUtils;
+import org.eclipse.lsp4e.LanguageServers;
+import org.eclipse.lsp4e.LanguageServersRegistry;
+import org.eclipse.lsp4j.DocumentSymbolParams;
+import org.eclipse.lsp4j.ServerCapabilities;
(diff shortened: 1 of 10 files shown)


vogella:
The three CodeQL alerts on `TestJSONResultLimitPreference` were false positives: `jsonDefinition` was read inside the lambda, the list was filled through the `servers::addAll` method reference, and `jsonServer` was passed to `countSymbols` twice. The CodeQL workflow analyses Java with `build-mode: none`, so it has no classpath and cannot resolve the LSP4E and LSP4J types, which makes every expression touching them opaque to it.

I still reworked the test rather than dismissing the alerts, since the code was more convoluted than it needed to be. Looking up the language server is now folded into a helper that returns it directly, so the three locals and the mutable list are gone and the test body is just two assertions around the preference change. It also filters nulls from `collectAll` explicitly instead of relying on the first element being the JSON server. The test still passes.

merks:
Generally it looks okay, but the long message looks a bit awkward and  the line break is in an awkward place:

image

Do we really need to mention "performance reasons" for the limit?

Here are other places with long text:

image

When it's broken to a new line, it's generally on a new sentence.

How about just keep it short?
```
JSonPreferencePage_maxItemsComputed=Maximum number of outline symbols and folding regions (0 = unlimited).
```

image

I rebased it hoping the ci build would build cleanly as it was fine for the PR to upgrade to Tycho 5.0.4.

merks:
@akurtakov 

I don't know why the pr build is failing.  Is this something we can just ignore?  I don't that the failure is related to any changes in this PR...

merks:
FYI, as an experiment I amended this commit with the shorter message, pushed it to my fork, and created a new PR:

https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/2198

merks:
Oh, master is aborting as well:

https://ci.eclipse.org/wildwebdeveloper/job/Wildwebdeveloper/job/master/1343/console

But this one was fine

https://ci.eclipse.org/wildwebdeveloper/job/Wildwebdeveloper/view/change-requests/job/PR-2197/

Maybe just flaky.   😕 

Oh, and this one is starting okay:

https://ci.eclipse.org/wildwebdeveloper/job/Wildwebdeveloper/job/PR-2198/1/console

merks:
Go figure.  Everything passed for this PR:

https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/2198

image

akurtakov:
@merks In the past Jenkins was not fetching custom images unless asked to fetch always but not it hits rate limits. Long story short - it needs the same fix as https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/commit/1ff27844d9b3693b32aa6f162f02d27f5cae46ee

merks:
Closed in favor of the same commit via the new PR.

https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper/pull/2198

